PR #41, opened 2022-06-28
allow dynamic color cases on icons

elevatebart:
The **icons** system before this Pull Request does not allow for hover, and focus colors to be different from the normal colors.

their PR adds the possibility to **prefix** the said colors with `hover`, `focus` and `hocus`. It also adds a way to make those pseudo-classes act with the parent group instead of the current icon.

### With a prefix

```html
<IconBook size="16" strokeColor="blue-600" hoverStrokeColor="jade-600" focusStrokeColor="jade-900" />
```

### With a prefix acting on a group

```html
<button class="group">
  <IconBook size="16" strokeColor="blue-600" hoverStrokeColor="jade-600" focusStrokeColor="jade-900" interactiveColorsOnGroup />Read
</button>
```


mapsandapps (approved):
looks good 😎

PR #47, opened 2022-07-20
fix: color of checkboxes

elevatebart:
Screen Shot 2022-07-20 at 2 06 30 PM

### DIFF
--- a/.changeset/quiet-eyes-begin.md
+++ b/.changeset/quiet-eyes-begin.md
@@ -1,0 +1,5 @@
+---
+'@cypress-design/css': patch
+---
+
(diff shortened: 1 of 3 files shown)


mapsandapps (approved):
this PR confused me at first. i see now that it's changing the disabled+checked state, which is an improvement. but the color of the enabled+checked state checkmark is still incorrect. it should be white. anyway, feel free to go ahead and merge this one.

elevatebart:
Did you yarn install (to rebuild the CSS plugin?)
The problem was mainly in the plugin as said in the changeset

PR #49, opened 2022-07-20
fix: avoid picking up class/className in icon color classes

elevatebart:
## Before

 `class` attribute (or `className` in React) would be wrongly picked up to build icon classes

before

# After

We use the white list of color attributes to filter the color props.

after


icon-registry/build-icons.mjs
```diff
@@ -199,6 +202,16 @@ async function generateIndex(iconsObjectUnique) {
     dedent`
   // THIS FILE IS AUTO GENERATED BY build-icons.mjs
 
+  /**
+   * All possible prop names for icon colors
+   */
+  export const ICON_COLOR_PROP_MANES = ${JSON.stringify(
```
mapsandapps, line 208: typo?

elevatebart: What do you mean?

mapsandapps: `MANES`?

elevatebart: wow, I need more sleep.
```suggestion
  export const ICON_COLOR_PROP_NAMES = ${JSON.stringify(
```

elevatebart: typo fixed

mapsandapps:
`strokeColor` is dark gray for all icons in [the story](https://cypress-design-h7hjqr4kk-cypress-io.vercel.app/?path=/docs/icons--page), regardless of the dropdown status. other uses of icons seem to be fine, though

elevatebart:
@mapsandapps good catch, I renamed a bunch of variables to get their names right but forgot one.

It should be working fine now.

mapsandapps (approved):

PR #57, opened 2022-07-22
fix: move the input for checkboxes outside of label

elevatebart:
For some screen readers, the label is not given properly when the input is inside.

A11y should have it outside.


mapsandapps (approved):
TIL

PR #58, opened 2022-07-26
feat: tooltip updates

elevatebart:
- Added an `interactive` prop so we can make the tooltip 'non-interactive' by default
- Made the tooltip arrow a cleaner design (no more backshadow)
- Added the `forcePlacement` prop to avoid the tooltip flipping when we do not want  it to


mapsandapps (commented):
it would be nice to add the `interactive` prop in the story. feel free to merge with or without that.

PR #70, opened 2022-08-08
feat: shared constants

elevatebart:
Make sharable constants a default for cypress components.

This change is only done for templates used by `yarn new:component`

It also shares cypress assertions by default so we run tests on both vue and react


mapsandapps (approved):
nice 😎

PR #92, opened 2022-10-06
fix: add 2 missing icons

elevatebart:


elevatebart:
@emilyrohrbough This is the icon you might need for the reporter I believe

icon-registry/icons/testing-type-component-solid_x24.svg
```diff
@@ -0,0 +1,11 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none"
```
mapsandapps, line 1: did you mean to leave the width and height in here?

icon-registry/icons/reporter-pending_x12.svg
```diff
@@ -0,0 +1,4 @@
+<svg viewBox="0 0 12 12" fill="none"
+  xmlns="http://www.w3.org/2000/svg">
+  <path d="M4.47023 2.3053C3.5245 2.69703 2.72821 3.44956 2.30552 4.47001C1.46015 6.51092 2.42932 8.85072 4.47023 9.69609C6.51114 10.5415 8.85094 9.57229 9.69631 7.53138C10.119 6.51092 10.088 5.41575 9.69631 4.47001C9.30458 3.52428 8.55205 2.72798 7.5316 2.3053" stroke="#BFC2D4" stroke-width="2" class="icon-dark"/>
```
mapsandapps, line 3: what is this icon for? how does it vary from status-pending-outline_x12? if it is a status icon, presumably we should use the existing status icon. if it is not a status icon, it should have `currentColor` as the stroke. 

either way, i'm not sure where the name for this icon is coming from. i don't see one with this name in the design system figma. it seems odd that the name of this icon would be "pending" if it's _not_ a status icon, and it seems odd that we would have an icon specific to the "reporter" in the design system.

icon-registry/icons/testing-type-component-solid_x24.svg
elevatebart: Good catch! I did not.

icon-registry/icons/reporter-pending_x12.svg
elevatebart: 3 solutions to Emily's need:
- We allow status icons to be colored (like they are in Figma) (my favorite)
- We create a whole set of status icons for dark mode
- We create duplicates of them in the design system (and give them weird names) (done here)

mapsandapps: i've come around to option 1 after seeing [the use case related to this PR](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/Cypress-App?node-id=4086%3A79235) 👍🏻

mapsandapps (approved):
PR could use a title change but is otherwise good to go!

mapsandapps (changes requested):
oops, changeset text needs an update too

mapsandapps (approved):

PR #101, opened 2022-10-17
Update the checkbox build

mapsandapps:


components/checkbox/react/Checkbox.tsx
```diff
@@ -77,7 +77,7 @@ export const Checkbox: FunctionComponent<CheckboxProps> = ({
         )}
         <span
           className={clsx([
-            `block border-1 rounded h-16px w-16px flex items-center text-white`,
```
elevatebart, line 80: Can you add the `// <windi-keep strokeColor="white"/>` here?

PR #103, opened 2022-10-18
Fix icon color issues

mapsandapps:
don't know how i managed to miss these 😕 

before:
Screen Shot 2022-10-18 at 11 50 47 AM

after:
Screen Shot 2022-10-18 at 11 50 54 AM


mapsandapps:
putting in draft b/c we might want to change the `noTests` icon, and if so, we might as well do that here 🤷🏻

elevatebart (approved):
nice

PR #108, opened 2022-10-20
fix: tooltip portal's z-index

elevatebart:


mapsandapps (commented):
missing a changeset but otherwise good to go 👍🏻

mapsandapps (approved):

PR #114, opened 2022-10-28
fix: add sizes to alerts

elevatebart:


mapsandapps (changes requested):
xs alert should be 32 px tall but is 36;
sm alert should be 40 px tall but is 44

also the text isn't vertically centered quite right in those sizes

elevatebart:
@mapsandapps should be fixed

mapsandapps (approved):

PR #116, opened 2022-12-08
Fix docs links for components with names that are not all lowercase

mapsandapps:


storybook/vue/stories/component-list.ts
```diff
@@ -1,3 +1,13 @@
 const comps = import.meta.globEager('../../../components/*/vue/package.json')
 
+export const capName = Object.keys(comps).reduce((acc, fullPath) => {
```
elevatebart, line 3: We should find a better name for this variable

`capitalizedComponentNameMap` ?

elevatebart (approved):
Apart from the single code style remark, It's good

PR #119, opened 2022-12-12
Adding size 16 browser and operating system icons

amehta265:
The cypress design system is currently missing browser and operating system icons.

These icons will be used in IATR on the app side: [Link to Figma](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/branch/QkCp2F69X0qyLo7qnYQ67e/Cypress-App?node-id=21467%3A202325&t=b8ojjS1LkkNmQN21-0)


elevatebart:
I think we should have them in here (it cost nothing and is useful everywhere).

I agree that we should remove all configuration from the icons since the color is part of the brand.

Thank you @amehta265

elevatebart:
@mapsandapps Both?

icon-registry/icons/os-generic_x16.svg
```diff
@@ -0,0 +1,10 @@
...
+</g>
+<defs>
+<clipPath id="clip0_4669_8471">
+<rect width="16" height="16" fill="currentColor"/>
```
elevatebart, line 7: this is the only one where it makes sense to have the classes. But do you really need this con for your dev ?

elevatebart (commented):
one class missing then I will add the changeset

icon-registry/icons/os-generic_x16.svg
amehta265: Yes I need it when there are multiple groups for a test. I would display this icon and the number of operating systems in the groups e.g. `<os-generic/> 2 operating systems`

PR #120, opened 2022-12-13
fix package json for vue-statusicon

elevatebart:


mapsandapps (approved):
let's go ahead and get this in, but why is the directory structure different for react vs. vue?

in react, it's ./dist/react/index.d.ts but
in vue, it's ./dist/StatusIcon/vue/index.d.ts 🤔

mapsandapps (changes requested):
needs a changeset

mapsandapps (approved):

PR #221, opened 2023-05-05
feat:  input component

jaimefps:
Creates Text Input component as [shown in Figma](https://www.figma.com/file/1WJ3GVQyMV5e7xVxPg3yID/Design-System%2C-v1.x---%40latest?type=design&node-id=851-9180&t=Blyzav1kotnWIy9v-0).

Screen Shot 2023-05-05 at 4 53 51 PM

Screen Shot 2023-05-05 at 4 53 58 PM


components/Input/react/Input.tsx
```diff
@@ -0,0 +1,125 @@
...
+  inputClasses,
+  StaticInputClasses,
+} from '@cypress-design/constants-input'
+import {
+  IconObjectMagnifyingGlass,
+  IconActionDeleteLarge,
+} from '@cypress-design/react-icon'
+
+export interface InputPropsJsx extends InputProps {
+  id: string
```
elevatebart, line 22: Why did you make id required for this field?

components/Input/react/Input.tsx
```diff
@@ -0,0 +1,125 @@
...
+      onBlur: disableFocusStyle,
+      onMouseLeave: disableFocusStyle,
+      onMouseOut: disableFocusStyle,
+    }
+  }, [enableFocusStyle, disableFocusStyle])
+  return {
+    propagateFocusProps,
+    hasFocus,
+  }
+}
```
elevatebart, line 49: Changing the style on focus is usually easier to build using tools like `focus-within`.

components/Input/assertions.ts
```diff
@@ -0,0 +1,62 @@
+/// <reference types="cypress" />
+
+export default function assertions(mountStory: (options?: any) => void): void {
+  it('renders disabled', () => {
+    mountStory({ disabled: true })
```
elevatebart, line 5: This is great !!!

components/Input/react/ReadMe.md
```diff
@@ -0,0 +1,94 @@
...
+```jsx live
+import {
+  default as Input,
+  VariantClassesTable,
+  SizeClassesTable,
+} from '@cypress-design/react-input'
+
+const Contrast = ({ children }) => (
+  <div className="bg-white p-[2px] rounded">{children}</div>
+)
```
elevatebart, line 30: Since we had to create this blob, my guts tell me that "by default" inputs should have a background. That's an opinion though.

Why did you decide to not add it?

elevatebart (commented):
Thank you @jaimefps this is a great start,

I took the liberty of adding the vue implementation.
I have a few questions that you could probably answser easily.

The typescript engine seems to work for me, I did not see your issues. Mybe it's a matter of restarting the TS server, maybe it's a `git clean` problem.

components/Input/react/Input.tsx
jaimefps: Is there an example of it in a react component at this time, it seems you've only done it in vue atm?

elevatebart: it is actually pure css... I'll modify this with an example.

PR #230, opened 2023-05-15
fix: prop button reactivity

### LINKED ISSUE #229: `vue-button` does not handle `disabled` responsively
`vue-button` does not toggle from a disabled to enabled state when the `disabled` prop is changed.

## Package/Version
`@cypress-design/vue-button@0.9.2`

## Repro test:

```ts
import DSButton from '@cypress-design/vue-button'
import { ref } from 'vue'

describe('<Button />', () => {
  it('responsively handles `disabled` changes', () => {
    const disabled = ref(true)

    cy.mount({
      render: () => (<div>
        <button data-cy="toggle" onClick={() => disabled.value = !disabled.value}>Toggle</button>

        <DSButton data-cy="ds-button" disabled={disabled.value}>DS Button</DSButton>
        <button data-cy="html-button" disabled={disabled.value}>Base HTML Button</button>
      </div>),
    })

    // HTML & DS Buttons start off disabled
    cy.findByTestId('html-button').should('be.disabled')
    cy.findByTestId('ds-button').should('be.disabled')

    // Click toggle button to update `ref` for `disabled` state
    cy.findByTestId('toggle').click()

    // Base HTML button properly enables
    cy.findByTestId('html-button').should('not.be.disabled')

    // FAILS - DS Button stays disabled
    cy.findByTestId('ds-button').should('not.be.disabled')
  })
})

```

elevatebart:
closes #229

### DIFF
--- a/.changeset/reactive-props.md
+++ b/.changeset/reactive-props.md
@@ -1,0 +1,5 @@
+---
+'@cypress-design/vue-button': patch
+---
+
+make props reactive
--- a/components/Button/vue/Button.vue
+++ b/components/Button/vue/Button.vue
@@ -53,6 +53,7 @@ export default defineComponent({
     return {
       componentTag: computed(() => (href ? 'a' : 'button')),
       href,
+      finalDisabled,
       buttonProps: reactive({
         ...attr,
         class: [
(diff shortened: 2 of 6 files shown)


elevatebart:
Thank you @mike-plummer, good eye. This fix will be delivered before (my?) lunch

PR #265, opened 2023-06-22
Stop Tooltip component from creating empty div

### LINKED ISSUE #260: Tooltip force open prop
Feature request:

It would be nice if the tooltip has an "open" or "show" prop to force it to display. 
This would be super helpful for development as well as mode dynamic UX ideas for showing tooltips programatically.

### LINKED ISSUE #263: Tooltip component creates empty div
I had been using the tooltip component - and per tooltip component used, it's creating an empty div

image

elevatebart:
closes #263
closes #260 

also add forceOpen for development and design

### DIFF
--- a/.changeset/forceopen-prop.md
+++ b/.changeset/forceopen-prop.md
@@ -1,0 +1,6 @@
+---
+'@cypress-design/react-tooltip': minor
+'@cypress-design/vue-tooltip': minor
+---
+
+add forceOpen prop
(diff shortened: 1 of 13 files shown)


jaimefps:
@elevatebart Seems that version 0.7.0 never was published? When I try to install it says that only 0.5.4 is available?

jaimefps:
Screen Shot 2023-07-18 at 1 58 55 PM

elevatebart:
I believe it is published.

https://www.npmjs.com/package/@cypress-design/react-tooltip?activeTab=versions

jaimefps:
> I believe it is published.
> 
> https://www.npmjs.com/package/@cypress-design/react-tooltip?activeTab=versions

My mistake was that I confused `@cypress/react-tooltip` and `@cypress-design/react-tooltip`, came back to it now and was able to install, thanks.

PR #284, opened 2023-07-18
Remove xl breakpoint on main

cbfrance:
Small issue with layout on the XL breakpoint around 1300px wide.

Before: 

Screenshot 2023-07-18 at 16 31 29

After:

Screenshot 2023-07-18 at 16 33 06

### DIFF
--- a/docs/.vitepress/theme/components/Layout.vue
+++ b/docs/.vitepress/theme/components/Layout.vue
@@ -170,7 +170,7 @@ const { frontmatter } = useData() as any
           @click="mobileMenuOpen = false"
         />
       </aside>
-      <main class="w-[800px] mx-[24px] md:mt-[24px]">
+      <main class="w-[800px] mx-[24px] min-[1420px]:mx-auto md:mt-[24px]">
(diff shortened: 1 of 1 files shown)


elevatebart (changes requested):
Maybe we should do a flex here to avoid this. The problem will now be on bigger screens.

docs/.vitepress/theme/components/Layout.vue
```diff
@@ -170,7 +170,7 @@ const { frontmatter } = useData() as any
           @click="mobileMenuOpen = false"
         />
       </aside>
-      <main class="w-[800px] mx-[24px] xl:mx-auto md:mt-[24px]">
+      <main class="w-[800px] mx-[24px] md:mt-[24px]">
```
elevatebart, line 173: How about this? this creates a custom breakpoint exactly where the problem would be. Check it out!

```suggestion
      <main class="w-[800px] mx-[24px] min-[1420px]:mx-auto md:mt-[24px]">
```

PR #288, opened 2023-07-24
fix: have accessibility check more than previously

### LINKED ISSUE #276: Accessibility Checks in `afterEach` always pass
We run axe checks in an `afterEach()` which means we get them automatically run at the end of every test, but it looks like in the this context axe doesn't run against the real DOM of the component.

To reproduce, check out [this branch](https://github.com/cypress-io/cypress-design/tree/marktnoonan/fail-axe-accordion) which modifies the Accordion to have violations and run [this component spec](https://github.com/cypress-io/cypress-design/blob/marktnoonan/fail-axe-accordion/components/Accordion/vue/AccordionVue.cy.tsx). 

Notice that the violations are only detected by the `cy.checkA11y()` that gets called from the `it` block, the other passes.

Screenshot 2023-07-11 at 10 00 17 AM

I'm not sure what the best solution is, other than to manually call checkA11y when the DOM is in a new state that needs to be tested, which is probably less often than at the end of every test.

elevatebart:
closes #276 

THis PR enables the WCAG 2 A and WCAG 2 AA rules. Previously, only rules specific to WCAG 2.1 AA were run. None of the rules under we run.


marktnoonan:
Oh nice, it was not some weird DOM problem in the afterEach - that was confusing to me since we don't clean up until before the next mount. This makes more sense!

PR #292, opened 2023-07-24
Allow Tab to accept data-pendo prop

mabela416:
The Cypress Cloud is using tabs and needs pendo tags to track events in full story and pendo


.changeset/data-pendo-prop-for-tab.md
```diff
@@ -0,0 +1,6 @@
+---
+"@cypress-design/react-tabs": patch
```
elevatebart, line 2:
```suggestion
"@cypress-design/vue-tabs": patch
"@cypress-design/react-tabs": patch
```

elevatebart (approved):

PR #306, opened 2023-08-18
Sidebar Navigation

cbfrance:
This branch contains a SidebarNavigation element and a LayoutNavigation to put it in. 

WIP

- Minimalist Cypress tests which merely verify that it renders. 
- Typed but probably not using correct conventions for typescript interfaces; need to refactor types.


components/Tabs/assertions.ts
```diff
@@ -14,7 +14,7 @@ export default function assertions(
     tabs: Tab[]
     activeId?: string
     variant?: keyof typeof variants
-  }) => void
```
elevatebart, line 17: to be removed?

test/react-app/index.html
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
```
elevatebart, line 1: ??

turbo.json
```diff
@@ -2,9 +2,7 @@
   "$schema": "https://turborepo.org/schema.json",
   "pipeline": {
     "build": {
-      "dependsOn": [
```
elevatebart, line 5: meh?

PR #310, opened 2023-08-24
Add-eslint-react-hooks

elevatebart:


components/Tabs/react/Tabs.tsx
```diff
@@ -82,7 +82,7 @@ export const Tabs: React.FC<TabsProps & React.HTMLProps<HTMLDivElement>> = ({
     window.addEventListener('resize', throttledUpdateActiveMarkerStyle)
     return () =>
       window.removeEventListener('resize', throttledUpdateActiveMarkerStyle)
-  }, [])
+  })
```
jaimefps, line 85:
```suggestion
  }, [throttledUpdateActiveMarkerStyle])
```

components/Tabs/react/Tabs.tsx
```diff
@@ -67,11 +67,11 @@ export const Tabs: React.FC<TabsProps & React.HTMLProps<HTMLDivElement>> = ({
...
     setMounted(true)
-  }
+  }, [activeId, tabs])
 
   React.useEffect(() => {
     updateActiveMarkerStyle()
-  }, [activeId])
+  }, [activeId, updateActiveMarkerStyle])
 
   const throttledUpdateActiveMarkerStyle = throttle(
```
jaimefps, line 76: might want to wrap this in a useMemo:

```
const throttledUpdateActiveMarkerStyle = useMemo(() => {
  return throttle( updateActiveMarkerStyle,
    100,
  )
},[updateActiveMarkerStyle])

```

components/DocMenu/react/_DocLink.tsx
```diff
@@ -23,7 +23,7 @@ export const DocLink: React.FC<DocLinkProps> = ({
     if (item.active) {
       onActive?.(activeLIRef?.current?.offsetTop || 0)
     }
-  }, [])
+  })
```
jaimefps, line 26: This will run on every rerender; your previous implementation was running only "on mount" (which is what happens with empty array in useEffect). 

So here you should keep the empty array; and you can eslint-ignore with a comment explaining it is meant to only happen on mount.

jaimefps: Or if you want to be HYPER kosher and avoid disabling the linter, then you can do this:
```
const dependenciesRef = React.useRef({ item, onActive, activeLIRef });
// Capture the current values on every render
dependenciesRef.current = { item, onActive, activeLIRef };

React.useEffect(() => {
  const { item, onActive, activeLIRef } = dependenciesRef.current;
  if (item.active) {
    onActive?.(activeLIRef?.current?.offsetTop || 0);
  }
  // Since dependencies are in a ref, you can still use an empty dependency array
}, []);
```

But a linter disabled with a comment might be good enough here.

components/Alert/react/Alert.tsx
```diff
@@ -107,37 +107,34 @@ export const Alert: React.FC<
...
 
   const [dismissed, setDismissed] = React.useState(false)
-  const [durationTimeout, setDurationTimeout] = React.useState<
-    number | undefined
-  >(undefined)
+  const durationTimeout = React.useRef<number | undefined>(undefined)
 
-  function clearDurationTimeout() {
+  const clearDurationTimeout = React.useCallback(() => {
     if (durationTimeout) {
```
jaimefps, line 113:
```suggestion
    if (durationTimeout.current) {
```

Otherwise this conditional will always pass then the output of useRef is `{ current: undefined }`

jaimefps: Actually, seems you can call `clearTimeout` on undefined, so you can remove the `if` altogether, and it's still fine

components/Alert/react/Alert.tsx
```diff
@@ -107,37 +107,34 @@ export const Alert: React.FC<
...
     setDismissed(true)
     onDismiss && onDismiss()
     clearDurationTimeout()
-  }
+  }, [clearDurationTimeout, setDismissed, onDismiss])
 
   React.useEffect(() => {
-    if (onDismiss && duration && !durationTimeout) {
-      setDismissed(false)
+    if (duration && !durationTimeout.current) {
```
jaimefps, line 126: `duration` is always truthy, since the output of useRef is `{current:undefined}`

components/Alert/react/Alert.tsx
```diff
@@ -107,37 +107,34 @@ export const Alert: React.FC<
...
       const timeout = setTimeout(dismiss, duration) as unknown as number
-      setDurationTimeout(timeout)
+      durationTimeout.current = timeout
     }
     return clearDurationTimeout
-  }, [])
+  })
```
jaimefps, line 131:
```suggestion
  },[])
```

If this is meant to happen only once (on mount) then you need an empty array and eslint-disable with a comment explaining this is meant to only run once.

components/Alert/react/Alert.tsx
```diff
@@ -107,37 +107,34 @@ export const Alert: React.FC<
...
 
   React.useEffect(() => {
     if (detailsRef.current && contentRef.current) {
       new DetailsAnimation(detailsRef.current, contentRef.current)
     }
-  }, [])
+  })
```
jaimefps, line 137: ditto on empty array and a comment if this was meant to run only once.

jaimefps (approved):

PR #343, opened 2023-11-15
Icon Improvements: ActionDelete & CheckmarkSolid

ryanjwilke:
This PR introduces:
- [x] Add more sizes for `IconActionDelete`
- [x] Fix the colors for the `IconCheckmarkSolid`

### DIFF
--- a/.changeset/rare-boats-talk.md
+++ b/.changeset/rare-boats-talk.md
@@ -1,0 +1,7 @@
+---
+'@cypress-design/icon-registry': patch
+'@cypress-design/react-icon': patch
+'@cypress-design/vue-icon': patch
+---
+
+Added more IconActionDelete sizes and fixed the colors for the IconCheckmarkSolid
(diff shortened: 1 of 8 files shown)


ryanjwilke:
@elevatebart It would appear this test is flaky, otherwise I'm not quite sure why it'd be failing: https://cloud.cypress.io/projects/89d3nq/runs/1472/overview?roarHideRunsWithDiffGroupsAndTags=1

Do we not have `retries` enabled? That would help resolve those issues quicker.

elevatebart:
I am going to enable retries on this test specifically. Since it is the flaky one. Hopefully when we get a11y through cypress directly, we can skip the flake.

elevatebart (approved):

elevatebart (changes requested):
The change on the checkmark icon is a breaking change. Should we do that change, we have to check if this icon is used elsewhere.

elevatebart (approved):
I checked on both properties and the icon does not seem to be used. 
We should be careful about this kind of "house-keeping" changes when they are not absolutely necessary.

ryanjwilke:
@elevatebart Yeah, I was curious about the checkmark changes too. However, the API for that icon was very backwards when actually working with it. Light colors were dark, and dark colors were light. It was really confusing.

PR #366, opened 2024-01-11
TestResult Component

ryanjwilke:
This PR creates a CSS-only solution for the `TestResult` components which are used throughout the App & Cloud.

- [x] @ryanjwilke to explore a CSS-only solution for test names
- [x] @ryanjwilke to build and test the Vue Component
- [x] @ryanjwilke to build and test the React Component
- [x] @elevatebart to ensure builds are passing in CI
- [x] @elevatebart to review component tests
- [x] @elevatebart to assist with component documentation
- [ ] @elevatebart to assist with testing components in ~App~ & Cloud

### Component Design
Screenshot 2024-01-22 at 20 18 35
Screenshot 2024-01-22 at 20 18 49


cypress/support/a11y-support.ts
```diff
@@ -1,27 +1,27 @@
 import 'cypress-axe'
 
-beforeEach(() => {
-  cy.injectAxe()
-})
+// beforeEach(() => {
```
ryanjwilke, line 3: Need to restore this before this PR is done.

package.json
```diff
@@ -106,7 +106,6 @@
     "vue-tsc": "^1.8.27"
   },
   "lint-staged": {
-    "*.{js,ts,jsx,tsx,vue}": "eslint --cache --fix",
```
ryanjwilke, line 109: Need to restore this when this PR is ready

elevatebart (approved):

cypress/support/a11y-support.ts
ryanjwilke: @elevatebart Just an FYI that we will need to turn this back on before this PR gets merged. I turned it off because it was just making noise for me while I was building everything out initially.

Eventually, I would love to be able to remove this entirely in favor of the a11y checks that are new built into the Cloud.

components/TestResult/vue/TestResult.vue
```diff
@@ -0,0 +1,100 @@
...
+} from '@cypress-design/constants-testresult'
+import { StatusIcon } from '@cypress-design/vue-statusicon'
+import {
+  IconChevronRightSmall,
+  IconStatusFlaky,
+  IconDocumentModifiedSquareDot,
+  IconDocumentAddedSquarePlus,
+} from '@cypress-design/vue-icon'
+
+defineProps<TestResultData>()
```
ryanjwilke, line 14: @elevatebart One thing I'm also remembering now is that we will probably need to add a `hover=true` property or something since the Debug page doesn't actually have a clickable row for these results.

ryanjwilke: This would conditionally add the classes needed for the hover state (i.e. filter, borders, transitions, etc)

components/TestResult/vue/TestResultVue.cy.tsx
```diff
@@ -0,0 +1,63 @@
...
+        />
+      </Button>
+    </TestResult>
+  )
+}
+
+describe('<TestResult/>', () => {
+  function mountStory() {
+    mount(() => (
+      <div class="p-[16px] antialiased">
```
ryanjwilke, line 53: @elevatebart Any chance there's a better way to handle the `antialiased` class in a more global way? It would be nice that we could make sure that we have some base styles or resets that can apply for every property using `cypress-design`. Otherwise, we would have to repeat some classes all throughout our DS components in order for things to render as expected downstream.

elevatebart: Done. Please check before I merge ;)

ryanjwilke: @elevatebart Thanks. Will this automatically kick in then in the component docs and component tests as well? This came up yesterday when Emil thought I was applying a medium font weight to the test names, but when I dug in, it was clear that this wasn't the case. In the end, it came down to the fact that we didn't have this anti-aliasing property applied across the board.

components/TestResult/vue/TestResult.vue
elevatebart: Should the prop be named "clickable" ?

components/TestResult/assertions.ts
```diff
@@ -0,0 +1,69 @@
...
+
+      cy.viewport(768, 400).log('@3xl')
+
+      cy.viewport(896, 400).log('@4xl')
+
+      cy.viewport(1024, 400).log('@5xl')
+    })
+
+    it('expand when groups are present', () => {
+      cy.get('[data-cy="cd-tr-container"]').eq(-1).as('container')
```
ryanjwilke, line 34: @elevatebart I could use your help on this one. It feels a bit dirty and brittle to use `eq()` here instead of passing through a property that better identifies each component. I was struggling to get that to work properly though locally when trying to use other properties like `id` or even a new data attribute like `data-cy-id` or something like that.

PR #471, opened 2024-08-01
Add icon upgrade square

estrada9166:
image


icon-registry/icons-static/action-upgrade-square_x16.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="currentColor" xmlns="http://www.w3.org/2000/svg" class="icon-dark">
```
elevatebart, line 1: You should have this class only on the path so the system can guess if strke or path shoudl be changed.

Also, you should not use a `fillColor` attribute here (though it will not be used in the system). It makes it harder to recognize which file is which icon in the explorer.

Traditionally we also remove the `width` and `height` attributes to get a better zoom on it in macos/windows/github

elevatebart (changes requested):
Just a few nits.

Also, as the PR auto-comment is mentionning you are missing a changeset file.

To add it `yarn changeset` and flag all 3 icon modules for a patch release.

TYTY

icon-registry/icons-static/action-upgrade-square_x16.svg
```diff
@@ -0,0 +1,3 @@
+<svg viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
```
elevatebart, line 1:
```suggestion
<svg viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" fill="transparent">
```

elevatebart (approved):

PR #486, opened 2024-08-28
Add New Flaky Icons

emilyrohrbough:
Update Flaky Icon to match latest designs


icon-registry/icons-static/status-flaky_x10.svg
elevatebart, line 1: These icons do not have a mean to be colored, say if we need to make them black and white or make them show lighter in a different context. Is this on purpose.

elevatebart (changes requested):
great PR overall.

Only one question: the icons don't accept colors. Is this on purpose? It can always be added later if need be.

Easier to add than to remove ;)

icon-registry/icons-static/status-flaky_x10.svg
emilyrohrbough: @elevatebart these are exported from Figma. I followed the pattern in place with the current icons

elevatebart (approved):

PR #497, opened 2024-10-28
Airplane icon added (angled)

emilmilanov:
Share icon is added
image


.changeset/fluffy-falcons-refuse.md
```diff
@@ -0,0 +1,7 @@
+---
+'@cypress-design/react-icon': minor
```
ryanjwilke, line 2: @emilmilanov I believe that the icon updates were considered as `patch` updates, instead of `minor` typically. We should probably continue to follow that pattern. Everything else looks good though.

ryanjwilke (changes requested):

.changeset/fluffy-falcons-refuse.md
warrensplayer: In looking at older PRs that introduce new icons, I saw both `minor` and `patch` used.

From a [semantic versioning](https://semver.org/) definition, I think this type of change would align with `minor`:  "when you add functionality in a backward compatible manner"

But I think you could argue either way.

emilmilanov: image

I used `yarn changeset` to create this part of the code; it doesn't offer a `patch` option, only major or minor. I can override this to patch if we agree on it. When I added icons to the DS before, I always used `minor`.

icon-registry/icons-static/action-share-large_x16.svg
warrensplayer, line 1: @emilmilanov Instead of just "share" as the name, would it make sense to call them "share-airplane" or something like that in case we ever add different icons for "share"?

ryanjwilke: I think I agree on that. I do think that the icons should probably just reflect the name of whatever it is that has been illustrated instead of a semantic name. (i.e. airplane, airplane-angled, etc)

.changeset/fluffy-falcons-refuse.md
warrensplayer: @ryanjwilke There is no limit to the numbers.

@emilmilanov I did go back and look a little bit harder and the CHANGELOG.md files and it does seem like most of the changes that add icons use `patch`.  So I guess update your changeset file to use `patch` as well for all three packages.

emilmilanov: Changeset is updated from `minor` to `patch`. I also renamed icons to object-airplane-angled...

ryanjwilke (approved):
@emilmilanov The code looks good. Do you mind just checking through the Percy screenshots as well? Most likely it's nothing and that build can be approved to before this merges in.

Let me know if it's largely flaky results. There's probably something we can do to reduce the flaky results in the future.

PR #503, opened 2024-11-18
add warning triangle icon

emilyrohrbough:
Screenshot 2024-11-18 at 2 34 18 PM


icon-registry/icons-static/warning-triangle_x16.svg
```diff
@@ -0,0 +1,7 @@
+<svg viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
ryanjwilke, line 1: @emilmilanov @emilyrohrbough The code for this warning trial icon seems oddly rather large for something that appears like a pretty simple icon. It might be worth seeing how this was constructed so it ends up being smaller (i.e. having a smaller footprint)

ryanjwilke:
The icon looks visually correct in the preview environment, but it might be nice to see if you can get the icon output to be a bit simpler.

PR #522, opened 2025-01-07
Adding square-minus icons

tbiethman:
Adding icon as seen in Figma here: https://www.figma.com/design/zVWa3fJtO0EWZ5FEhqtCRL/Branch-Review%2C-v1.1---%40latest?t=xcmiNHxq0O0NXmJc-0

Wasn't sure if bundling the red color by default would be expected/desirable - happy to tweak this if so.

Screen Shot 2025-01-07 at 5 00 15 PM

### DIFF
--- a/.changeset/mighty-snakes-thank.md
+++ b/.changeset/mighty-snakes-thank.md
@@ -1,0 +1,7 @@
+---
+'@cypress-design/react-icon': minor
+'@cypress-design/vue-icon': minor
+'@cypress-design/icon-registry': minor
+---
+
+Adding document-square-minus_x12 icon
--- a/.changeset/mighty-snakes-thank.md
+++ b/.changeset/mighty-snakes-thank.md
@@ -4,4 +4,4 @@
 '@cypress-design/icon-registry': minor
 ---
 
-Adding document-square-minus_x12 icon
+Adding square-minus_x12, square-minus_x16, and square-minus_x24 icons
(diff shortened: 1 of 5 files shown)


.changeset/mighty-snakes-thank.md
```diff
@@ -0,0 +1,7 @@
+---
+'@cypress-design/react-icon': minor
+'@cypress-design/vue-icon': minor
+'@cypress-design/icon-registry': minor
```
emilyrohrbough, line 4: these are typically patch releases

emilyrohrbough (approved):

.changeset/mighty-snakes-thank.md
tbiethman: I was referencing https://github.com/cypress-io/cypress-design?tab=readme-ov-file#adding-a-new-icon - imo minor makes sense for net new (non-fix) additions?

emilyrohrbough: I tend to agree with minor as well, just the pattern I've seen in this repo for new icons

icon-registry/icons-static/document-square-minus_x12.svg
```diff
@@ -0,0 +1,4 @@
+<svg viewBox="0 0 12 12" fill="none" xmlns="http://www.w3.org/2000/svg">
```
ryanjwilke, line 1: @emilmilanov @tbiethman Do you know where this icon came from? I'm second guessing the name because it doesn't really look like a "document", but I can't find it in the Figma DS either to verify what the name of this icon is there.

tbiethman: I didn't see an explicit name in [Figma](https://www.figma.com/design/zVWa3fJtO0EWZ5FEhqtCRL/Branch-Review%2C-v1.1---%40latest?node-id=3072-206228&t=uYjIG5f1CwjhBdVJ-0) so I named it similarly to the existing square added icon, `document-added-square-plus`. I might have put the cart before the horse if it's not in the Figma DS at all though.

tbiethman: Based on Emil's feedback I'll update this to `modified_square_deleted` and add the additional sizes that are DS for this.

ryanjwilke (approved):

PR #525, opened 2025-01-28
add regenerate config icon

mabela416:


ryanjwilke (approved):
Just double-check the changeset on the previous icon PRs as this one feels like it's missing something.

PR #535, opened 2025-03-04
Add overlay arrow icons

chrisbreiding:
Adds overlay arrow icons for use on popovers/dialogs like the [test generation feedback comment overlay](https://www.figma.com/design/mHHSTgevvYXSYQgLZhJvPY/UI-Coverage-%3A%3A-Generate-tests?node-id=4417-237290&t=tAeDCogFGH6tO3YP-4).

Screenshot 2025-03-04 at 11 24 28 AM


ryanjwilke (changes requested):
I would suggest looking into how the [Tooltips](https://design.cypress.io/components/vue/Tooltip.html) components were built as those had to solve the same problem. However, I don't think they needed to add these arrows as an `Icon` in order to accomplish this style.

From my perspectives, the `Icon` component makes more sense for things that we would expect to be re-used in many different places, but this svg shape feels a bit niche and not something that's re-usable in the same way as the other icons.

I might suggest just adding the popover to the DS and shipping that as a component for others to use (similar to the Tooltip component)

chrisbreiding:
Going to go in a different direction for this

PR #550, opened 2025-04-09
Adding gray-1100

emilmilanov:
Introducing darker gray coloro mean for the backgroubd of Test Replay and Cypress Studio.

image


ryanjwilke (approved):
The code seems fine to me. I can't say that I've had to update the colors before, but if there's any components using the new colors, my guess is that we'd need to release new versions of those components as well.

If not, then I think this should be good to go 👍

PR #552, opened 2025-04-09
docs: Update the button component install instructions

jennifer-shehane:
I couldn't figure out why the types were not being picked up properly for my react-buttons component. It turns out those types are being pulled in from the constants-button dependency and if you don't have that package also installed - it just cannot access it. 

This updates the instructions to tell you to install the constants-button package also and why you'd want it installed.


ryanjwilke (approved):
Looks good to me 👍

PR #553, opened 2025-04-10
Adding red-dark-mode and outline-red-dark-mode buttons

emilmilanov:
Preview:
image
image

Buttons are previewed on a dark background which indicates that they are dark mode buttons.

Currently we're adding these buttons as new button variants.

In the future we will introduce prop `theme` with values inherit, light and dark. This will enable us to decrease number of variants, but will also require refactoring.

### DIFF
--- a/.changeset/clean-plants-change.md
+++ b/.changeset/clean-plants-change.md
@@ -1,0 +1,7 @@
+---
+'@cypress-design/constants-button': minor
+'@cypress-design/react-button': minor
+'@cypress-design/vue-button': minor
+---
+
+Adding outline-red-dark mode and red-dark-mode buttons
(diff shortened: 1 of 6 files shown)


.changeset/clean-plants-change.md
```diff
@@ -0,0 +1,7 @@
+---
+'@cypress-design/constants-button': minor
+'@cypress-design/react-button': minor
+'@cypress-design/vue-button': minor
+---
+
+Adding outline-red-dark mode and red-dark-mode buttons
```
ryanjwilke, line 7: @emilmilanov You may want to add notes for both `red` and `jade`

ryanjwilke (approved):
Code looks good. I'd recommend just improving the changeset notes to account for all changes now.

PR #556, opened 2025-04-11
Update buttons to use default variants for outline-, white, and dark-mode buttons

jennifer-shehane:


components/Button/react/Button.tsx
```diff
@@ -32,12 +32,17 @@ export const Button: React.FC<ReactButtonProps> = ({
...
   ...rest
 }) => {
-  const finalVariant =
-    disabled && !['outline-dark', 'outline-light', 'link'].includes(variant)
-      ? variant.includes('outline') || variant === 'white'
-        ? 'outline-disabled'
-        : 'disabled'
-      : variant
+  // Determine if we should keep the original variant
+  const shouldKeepOriginalVariant =
```
karlsnyder0, line 36: This seems confusing. Suggestion-

```
  // Determine if we should keep the original variant
  const useDisabledVariant =
    // If not disabled, always keep original
    disabled &&
    // Keep original for dark mode variants
    !variant.includes('dark-mode') &&
    !variant.includes('outline') &&
    variant !== 'white'
```

karlsnyder0 (approved):

PR #571, opened 2025-05-05
Adding tag, gray900

emilmilanov:
image

### DIFF
--- a/components/Tag/constants/src/index.ts
+++ b/components/Tag/constants/src/index.ts
@@ -12,37 +12,55 @@ export const CssColor = {
   white: {
     default: 'bg-white text-gray-700 border-gray-100',
     dark: 'bg-gray-50 text-gray-700 border-gray-50',
+    'default-outline': 'bg-white text-gray-700 border-gray-200',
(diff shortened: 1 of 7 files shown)


components/Tag/constants/src/index.ts
```diff
@@ -17,6 +17,10 @@ export const CssColor = {
     default: 'bg-gray-50 text-gray-700 border-gray-100',
     dark: 'bg-gray-700 text-white border-gray-700',
   },
+  gray900: {
```
ryanjwilke, line 20: @emilmilanov I would recommend calling this one `gray-dark` maybe just to align closer to how we call our `Button` components.

emilmilanov: I agree.

emilmilanov:
PR is ready for a review @ryanjwilke 

I updated all tag variants to support outline.

The only issue I see is that `white dark` and `gray light` variants look the same. This issue existed before and is consequence of the decision that each color variant has light and dark option.

ryanjwilke (approved):
This seems fine for now. I don't have any major concerns on my end.

PR #582, opened 2025-05-16
add redo and undo icons

mabela416:


ryanjwilke (approved):
Code and preview environment looks good to me. You may want to check out the Percy build as well. Not sure why there or so many visual diffs in this PR.

PR #585, opened 2025-05-20
feat: adding technology-browsers-light_x16

tbiethman:
Source: https://www.figma.com/design/1WJ3GVQyMV5e7xVxPg3yID/Design-System?node-id=6814-10841&t=ImlHgGvF3L3ERamM-0


icon-registry/icons-static/technology-browsers-light_x16.svg
```diff
@@ -0,0 +1,21 @@
+<svg
+  viewBox="0 0 16 16"
+  fill="none"
+  xmlns="http://www.w3.org/2000/svg"
+>
+  <path
+    d="M6 7C6 6.44772 6.44772 6 7 6H13C13.5523 6 14 6.44772 14 7V9H6V7Z"
+    fill="#E1E3ED"
```
karlsnyder0, line 8: I think this fill needs currentColor

tbiethman: The first two paths are the "toolbars" of the browsers, with the third being the outlines for the two "windows". Making them all currentColor (rather than just the outline, like I have here) would result in the toolbars matching the outline and, by default, looking like just a thick top border.

Screenshot 2025-05-20 at 9 37 12 PM
 
 vs.
 
Screenshot 2025-05-20 at 9 38 01 PM

With the icon-* classes those colors can be overridden in either case, so either way could produce something workable. But I think a non-currentColor default for these makes sense?

karlsnyder0: Which version does your PR produce? The first?

karlsnyder0 (approved):
I "think" this is correct, however it doesn't match with the icon instructions. I've seen this before and I was wondering why there were other fills in icons in the repo.

icon-registry/icons-static/technology-browsers-light_x16.svg
tbiethman: Indeed, the current state produces the first. The [preview](https://cypress-design-git-browsers-icon-cypress-io.vercel.app/Icons.html) shows the strokeColor/fillColor props working as I'd expect, which at the end of the day I think is the most critical piece here.

PR #592, opened 2025-05-29
add colors css stylesheet for color variables

mabela416:
Currently we can only use the design system colors if we import them like this in our javascript files 
```
import { colors } from '@cypress-design/css'
```

but when we're working with css/scss stylesheets, we can't import the colors. In the dashboard, we have a separate design system where we defined the colors ourselves and that's what we've been using but for studio, we're not using this cypress services design system so I decided to add a stylesheet to the package `@cypress-design/css` so that we can import it in our css/scss files like this

```
@use "@cypress-design/css/dist/index.cjs.css" as *;
```

Like this we'll also have a central place where all our color definitions exist.


ryanjwilke:
@mabela416 Any idea why Percy is missing so many screenshots? It seems a bit odd. I don't remember seeing this issue in other PRs but I could be wrong. I know we've struggled with flake in this repo in the past.

mabela416:
> @mabela416 Any idea why Percy is missing so many screenshots? It seems a bit odd. I don't remember seeing this issue in other PRs but I could be wrong. I know we've struggled with flake in this repo in the past.

I've seen this in other PRs. Not sure why because the other PRs I've seen this happen to are just for icons so we're not making big changes so I don't think the missing snapshots are related to my work

ryanjwilke:
@mabela416 That seems like a better compromise for sure. The only part that's hard to know is what I'm importing from that file. Is it just colors? Is it the entire design system and other selectors?

Again, your call, but I think it can be helpful to make sure the API as simple and understandable as possible for future engineers.

mabela416:
> @mabela416 That seems like a better compromise for sure. The only part that's hard to know is what I'm importing from that file. Is it just colors? Is it the entire design system and other selectors?
> 
> Again, your call, but I think it can be helpful to make sure the API as simple and understandable as possible for future engineers.

Yeah this is limited to the `cypress-design/css` package and right now it's just what's in `colors.css` because I'm importing that file [here](https://github.com/cypress-io/cypress-design/pull/592/files#diff-94498775a3de1496a56e77e8260aedeb2ef0ffcb9dbf231adf82d067c35938c7R2). Then, when we bundle the code all the css we imported in any of the `cypress-design/css` files get bundled into one file - `index.css`. If I create other css styles and import them to any of my files in the `cypress-design/css` package, then they'll also be included in the output file `index.css`.

mabela416:
> > @mabela416 That seems like a better compromise for sure. The only part that's hard to know is what I'm importing from that file. Is it just colors? Is it the entire design system and other selectors?
> > Again, your call, but I think it can be helpful to make sure the API as simple and understandable as possible for future engineers.
> 
> Yeah this is limited to the `cypress-design/css` package and right now it's just what's in `colors.css` because I'm importing that file [here](https://github.com/cypress-io/cypress-design/pull/592/files#diff-94498775a3de1496a56e77e8260aedeb2ef0ffcb9dbf231adf82d067c35938c7R2). Then, when we bundle the code all the css we imported in any of the `cypress-design/css` files get bundled into one file - `index.css`. If I create other css styles and import them to any of my files in the `cypress-design/css` package, then they'll also be included in the output file `index.css`.

I [add more info to ReadMe](https://github.com/cypress-io/cypress-design/pull/592/commits/daa99bac663fa5ef59e6b7be98a4a08ad01d1929) the ReadMe to add a note about this

PR #604, opened 2025-06-23
Adding AI disabled icon in 16x size

emilmilanov:
image


icon-registry/icons-static/general-sparkle-single-large-disabled_x16.svg
```diff
@@ -0,0 +1,8 @@
...
+  <path d="M2 14L14 2" stroke="#E45770" stroke-linecap="round"/>
+</svg>
+
```
ryanjwilke, line 6: @emilmilanov There's a lot of line breaks at the end of this file for some reason. Typically this is something a linter would catch, so just something to look into.

PR #667, opened 2026-05-27
docs(Select): stage 1 — instructions + architecture for new Select component

emilmilanov:
## Summary

Stage 1 of 4 for porting cypress-services' `SingleSelect` (built on `react-select`) into cypress-design as a new `Select` component. This PR is **docs only** — it lays out the public API, item shapes, states, accessibility model, and implementation notes so reviewers can sign off on the surface before any code is written.

- `components/Select/instructions.md` — consumer-facing usage doc (props, events, item shapes, states, accessibility, known limitations).
- `components/Select/architecture.md` — implementation notes (file layout, composition over invention, state ownership, constants keying, popover positioning, filtering, click-outside/ESC, trigger swap, extension points).
- `.agents/index.md` — registers `Select` in the component list.

## Stacked-PR series

This PR is the base of a four-PR stack. Each subsequent PR bases on the previous and merges together when the whole component is ready.

1. **`select-instructions` → `main` (this PR)** — docs only.
2. `select-component` → `select-instructions` — constants, React + Vue components, demo.
3. `select-tests` → `select-component` — Cypress component tests, Percy snapshots.
4. `select-accessibility` → `select-tests` — final ARIA + keyboard polish.

## Key API decisions captured in the docs

- **Built natively** (no `react-select`). Composes existing primitives: Textbox (trigger + search), Tabs (header), Checkbox (`checkbox` row), Button (`button` row), Tag, Icon.
- **Hybrid row API** — typed `type` field (`default | headline | divider | checkbox | user | button | custom`) with a `custom.render` escape hatch.
- **Single-select v1.** Multi-select reserved for v2 via a future `multiple` prop.
- **Default Textbox trigger + `trigger` render-prop / slot escape hatch.**
- **Header / footer** — slots (`#footer`, `#trigger`) plus structured shorthands (`headerTitle`, `headerTabs`, `searchable`, `footerLabel`, `footerAction`).
- **Alignment** — `align: 'left' | 'right'`. No collision detection (out of scope for v1).
- **No `radio` row in v1** — blocked on a shared Radio component landing in cypress-design.
- **ARIA combobox + listbox pattern**, focus tracked via `aria-activedescendant` so the search Textbox keeps real DOM focus while arrows traverse rows.

## Test plan

[...]

PR #668, opened 2026-05-27
feat(Select): stage 2 — implement Select component (React + Vue)

emilmilanov:
## Summary

Stage 2 of 4 — the actual Select implementation. Three new workspace packages, full Figma scope, no `react-select` dependency, React + Vue parity.

Stacked on top of [cypress-design#667](https://github.com/cypress-io/cypress-design/pull/667) (`select-instructions`). When that merges, GitHub auto-rebases this PR's base to `main`.

## What ships

- **`@cypress-design/constants-select`** — class tables, `SelectItem` discriminated union, defaults, sizing helpers (`toCssLength`, `buildSizingStyle`).
- **`@cypress-design/react-select`** — `Select.tsx`, `SelectOptionList.tsx`, `SelectOptionItem.tsx`, plus `filter-items.ts`.
- **`@cypress-design/vue-select`** — `Select.vue`, `_SelectOptionList.vue`, `_SelectOptionItem.vue`, plus `filter-items.ts`.
- **`docs/src/demos/Select.vue`** — live demo covering every variant, content type, theme, alignment, and the header/footer flavor.

## Features

- **Default Button trigger** with rotating chevron (`IconChevronDownSmall`). Replaceable via `trigger` render-prop (React) / `#trigger` slot (Vue).
- **Row content types**: `default`, `headline`, `divider`, `checkbox`, `user`, `button`, `custom`. `radio` deferred until cypress-design ships a Radio primitive.
- **Header**: `headerTitle`, `headerTabs` (uses cypress-design `Tabs`), `searchable` (uses cypress-design `Textbox`). Search filters by case-insensitive substring on `label` and drops orphaned headlines.
- **Footer**: `footerLabel` + `footerAction` shorthand, or arbitrary footer slot.
- **Themes**: `light`, `dark` — colors extracted explicitly from Figma; no auto-mapping.
- **Sizes**: `32`, `40`.
- **Alignment**: `align: 'left' | 'right'`.
- **Popover sizing**: `width`, `minWidth`, `maxWidth`, `height`, `maxHeight` — inline-style on the panel only, with header / footer pinned and items area scrolling.
- **Single-select selection**: controlled (`value` + `onChange`) and uncontrolled (`defaultValue`).
- **Click-outside**: capture-phase `mousedown` listener on `document`. `Escape` closes and refocuses trigger. `Tab` closes.
- **Keyboard**: `ArrowUp` / `ArrowDown` / `Enter` / `Space` traverse + select. Structural rows (`headline`, `divider`, `button`, disabled) are skipped.
- **ARIA**: `role=\"combobox\"` on trigger, `role=\"listbox\"` on panel, `role=\"option\"` + `aria-selected` on rows, `aria-controls`, `aria-expanded`, `aria-activedescendant`, `aria-disabled`. Full polish lands in Stage 4.

## API change vs. Stage 1 docs

Stage 1 had a `variant` prop forwarded to the (then Textbox) trigger. Now that the trigger is a Button, the visual style is controlled via `triggerVariant` (any Button variant string) and `variant` is gone. Instructions.md updated accordingly.

## Vue-specific note (Boolean prop coercion)

Vue normalizes absent Boolean props to `false`, so we can't distinguish "consumer didn't pass `open`" from "consumer passed `open=false`". v1 uses internal state as the source of truth: `open` and `defaultOpen` both seed the initial value, `update:open` fires on every change. Full controlled-mode for the open state is React-only for now. This is documented inline in `Select.vue` and called out as a known limitation in `instructions.md`.

## Verification

`yarn dev` and visited `/components/Select`. Verified by browser-driving the demo:

- Triggers render in both themes, both sizes, all variants, and the disabled state.
- Click opens the popover with chevron rotated. Click an option → trigger label updates, popover closes, selected row highlighted on next open.
- Click outside closes (via mousedown capture). The complex popover (`headerTitle` + `headerTabs` + `searchable` + `footerLabel` + `footerAction`) renders Tabs, search Textbox, all items, and footer correctly.
- Dark popover variant uses Figma dark colors (`gray-1000` background, `gray-800` border, `indigo-500/20` selected).
- ARIA: `aria-haspopup=\"listbox\"`, `aria-expanded` toggles, `aria-controls` points at the listbox id, options have `role=\"option\"` and `aria-selected`.

React side is implementation-complete and structurally analogous; it ships unverified visually since the live docs demo only renders Vue. Cypress component tests in Stage 3 will cover it.

## Test plan

- [ ] Pull the branch and run `yarn dev`. Visit `/components/Select`. All 14 demo dropdowns should render and open.
- [ ] Spot-check selection: open the first dropdown, click an option, confirm the trigger updates.
- [ ] Spot-check keyboard: tab to the first trigger, hit `Enter` / `ArrowDown`, navigate with arrows, `Enter` to select, `Escape` to close.
- [ ] Spot-check dark mode: scroll to the dark section, open any dropdown, confirm dark colors.
- [ ] Confirm the API surface (props on `<Select>`, item shapes, slots) matches the Stage 1 docs after the `triggerVariant`-replaces-`variant` change.

---

> [!NOTE]
> **Medium Risk**
> Large new public surface with non-trivial keyboard, controlled-value, and accessibility behavior; additive to the design system with no auth or data-path changes.
> 
> **Overview**
> Adds a **new single-select dropdown** to cypress-design as three workspace packages: **`@cypress-design/constants-select`** (types, Tailwind class tables, search helpers), **`@cypress-design/react-select`**, and **`@cypress-design/vue-select`**. The default export is full **`Select`** (Button trigger + popover); **`SelectOptionList`** and **`SelectOptionItem`** are also exported for inline panels.
> 
> Implementation composes existing primitives (Button, Tabs, Checkbox, Textbox, Tag, Icon) with Figma-driven light/dark styling, sizes **32** / **40**, optional header (title row, back button, tabs, search), footer, and popover sizing. Row types include **default**, **headline**, **divider**, **checkbox** (toggle-off emits **`undefined`**), **user**, **button**, and **custom**. Search filtering lives on the root component so keyboard navigation and **`aria-activedescendant`** stay aligned with the filtered list; **`searchFilters`** can disable filtering while keeping the search UI.
> 
> **`instructions.md`** is updated for **`triggerVariant`** (replacing **`variant`**), default size **32**, richer header props, **`iconRight`**, checkbox clear behavior, and button + listbox ARIA (not combobox). Docs add **`ReadMe.md`**, package readmes, and a broad **`docs/src/demos/Select.vue`** showcase.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a64d9635186f36494b524afaa6857d543eb2d251. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


components/Select/react/Select.tsx
```diff
@@ -0,0 +1,352 @@
...
+    }
+    if (e.key === 'Enter') {
+      e.preventDefault()
+      // No-op when nothing is focused yet (user hasn't pressed an arrow).
+      if (focusedIndex === -1) return
+      const realIndex = selectableIndices[focusedIndex]
+      if (realIndex === undefined) return
+      const item = items[realIndex]
+      handleSelect(item)
+    }
```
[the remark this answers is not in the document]
emilmilanov: Fixed in 7f69e0c5. Updated `instructions.md` to drop the misleading 'Enter fires `onClick` on `button` rows' line — button rows are intentionally excluded from keyboard traversal in v1, so the docs were promising behavior the code didn't deliver. Keyboard reach for button rows is tracked for the accessibility branch (stage 4).

components/Select/vue/Select.vue
```diff
@@ -0,0 +1,334 @@
...
+    : internalValue.value,
+)
+watch(
+  () => [props.value, props.modelValue],
+  () => {
+    if (props.value !== undefined) internalValue.value = props.value
+    else if (props.modelValue !== undefined)
+      internalValue.value = props.modelValue
+  },
+)
```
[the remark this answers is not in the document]
emilmilanov: Fixed in 7f69e0c5 (both Vue and React, same shape of bug). `isValueControlled` is now sticky — locked at first observation and never flips back to 'uncontrolled' mid-session. The Vue prop watcher also syncs `internalValue` to `undefined` so any fallback path stays in sync after a checkbox toggle-off.

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,361 @@
...
+  // would keep displaying the cleared selection.
+  const isValueControlledRef = React.useRef<boolean | null>(null)
+  if (isValueControlledRef.current === null) {
+    isValueControlledRef.current = valueProp !== undefined
+  }
+  const isValueControlled = isValueControlledRef.current
+  const [internalValue, setInternalValue] = React.useState<string | undefined>(
+    defaultValue,
+  )
+  const value = isValueControlled ? valueProp : internalValue
```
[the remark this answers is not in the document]
emilmilanov: Good catch — fixed in 3767d244 (both React and Vue). Refactored to drop the controlled/uncontrolled boolean entirely: every booleanly approach had a broken case (sticky-on-first-defined breaks lazy controlled state; dynamic flips on toggle-off). Instead, `internalValue` mirrors the latest committed selection and the prop wins via `valueProp ?? internalValue` (React) / `props.value ?? props.modelValue ?? internalValue` (Vue). Verified live: a Select with `useState<string|undefined>(undefined)` + `value={x}` now correctly tracks parent updates after the initial render, and checkbox toggle-off still resets the trigger to the placeholder.

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,369 @@
...
+  //     to a stale internal value).
+  // Instead, `internalValue` mirrors the latest *committed* selection and
+  // the prop wins whenever it's defined. The `??` chain lets both modes
+  // coexist: controlled consumers see `valueProp` through; uncontrolled
+  // consumers see `internalValue` through; toggle-off lands `undefined`
+  // in both and the trigger correctly shows the placeholder.
+  const [internalValue, setInternalValue] = React.useState<string | undefined>(
+    defaultValue,
+  )
+  const value = valueProp ?? internalValue
```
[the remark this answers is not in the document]
emilmilanov: Fixed in a658866a (both React and Vue). Switched to a sticky-toward-controlled discriminator: a `wasControlled` ref/state flips true the first time the component observes a defined `value` (or `modelValue` in Vue) and stays true for the rest of its life. Once controlled, the prop wins directly, so a parent clearing `value` to `undefined` now shows the placeholder instead of falling back to the last clicked row. Verified live: lazy v-model (initial `undefined` → user picks Alpha) → trigger shows 'Alpha'; controlled v-model → checkbox toggle-off → trigger drops back to placeholder. Both cases work.

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,375 @@
...
+  // Open state
+  defaultOpen?: boolean
+  open?: boolean
+  onOpenChange?: (open: boolean) => void
+
+  // Escape hatches
+  trigger?: React.ReactNode | ((ctx: SelectTriggerContext) => React.ReactNode)
+  className?: string
+  popoverClassName?: string
+  id?: string
```
mabela416: Some of these props are also in`SelectOptionList` and I think `SelectOptionItem`. We can create interfaces or types to share between these 3 components

components/Select/react/SelectOptionList.tsx
```diff
@@ -0,0 +1,355 @@
...
+                  square
+                  variant={theme === 'dark' ? 'outline-dark' : 'white'}
+                  aria-label={headerButton.ariaLabel}
+                  onClick={headerButton.onClick}
+                  className={SelectConstants.CssHeaderBackButtonSpacingClasses}
+                >
+                  {React.createElement(headerButton.iconLeft, {
+                    size: '16',
+                    interactiveColorsOnGroup: true,
+                  })}
```
mabela416: I'm curious why we need to call `React.createElement`, I would expect the client to pass an icon from the design system which would already be a React component

components/Select/react/filter-items.ts
```diff
@@ -0,0 +1,56 @@
...
+  // First pass: mark every item as kept or dropped.
+  const keep = items.map((item) => {
+    if (item.type === 'headline' || item.type === 'divider') return true
+    if (item.type === 'button') return true
+    const label = getItemLabel(item)
+    if (!label) return true
+    return label.toLowerCase().includes(trimmed)
+  })
+
+  // Second pass: drop headlines that have no non-structural kept item before
```
mabela416: what does no non-structural kept item mean here?

mabela416 (commented):
Would it be possible to have the component rendered next to the code kind of like what Ryan shared from the Atlassian design system?

The demo is also only written in vue but most of the time we'll be using the React one so it'd be nice to either have the demo written in both React and Vue or in React instead of Vue

components/Select/react/Select.tsx
emilmilanov: Good point — these three components do share a lot of prop shape (theme, size, items, value, the whole `headerTitle`/`headerButton`/`headerIconLeft`/`headerTag`/`headerIconRight` group, `searchable`/`searchPlaceholder`/`searchFilters`, `footerLabel`/`footerAction`, and the sizing fields).

I'm not sure which approach you'd prefer here — could you guide me on which of these you'd like?

1. **Pick/Omit from one source.** Make `SelectOptionListProps` the source of truth in `@cypress-design/constants-select` and derive the others, e.g. `type SelectProps = Omit<SelectOptionListProps, 'onSelect' | 'focusedIndex'> & { modelValue?: string; defaultOpen?: boolean; open?: boolean; ... }`. Single place to edit; JSDoc on the header props lives in one file.

2. **Small named prop groups.** Extract `SelectHeaderProps`, `SelectSizingProps`, `SelectFooterProps` in `constants-select/src/index.ts` and have all three components `extends`/spread them. More composable but spreads the docs across a few interfaces.

3. **Compositional `extends`.** `interface SelectProps extends SelectOptionListProps { ... }` with the list-only fields omitted. Closest to plain inheritance, simplest to read.

I don't have a strong preference and the repo doesn't have an existing convention for this (the other components don't have inner sub-components like Select does), so I'd appreciate your call on which direction fits the codebase best.

components/Select/react/SelectOptionList.tsx
emilmilanov: Good catch — we don't need `React.createElement` here. The issue was that the icon is passed as a *component reference* (`React.ComponentType<...>`), and JSX requires a capitalized identifier to render a component variable, which is why I had reached for `createElement` as the escape hatch.

Three options I considered:

1. **Local PascalCase alias (chosen).** Rename the destructured icon props to PascalCase (`HeaderButton`, `HeaderIconLeft`, `HeaderIconRight`) so they render as regular JSX (`<HeaderIconLeft size="16" interactiveColorsOnGroup />`). Keeps the existing API — consumers still pass `IconArrowLeft` (a component reference) — and the component continues to inject `size`/`interactiveColorsOnGroup`/`className` consistently across all header icons.
2. **Change the prop type to `ReactNode`** so consumers pass `<IconArrowLeft size="16" />` directly. Cleanest call site, but we lose the ability to enforce sizing and interactive-colors from inside the component, and it would diverge from the Vue API (which uses `<component :is="...">`-style passing).
3. **Accept `ReactElement` + `React.cloneElement`** to inject props. Generally considered an anti-pattern (cloneElement silently overrides consumer-provided props).

Went with option 1 — answers the `createElement` smell directly, keeps the React/Vue APIs symmetric, and preserves the design-system invariant that header icons render at the right size with interactive colors. Diff is ~12 lines in `SelectOptionList.tsx`, no consumer-facing change.

components/Select/react/filter-items.ts
emilmilanov: Good call @mabela416 — that wording was hard to parse (double-negative + an undefined "structural" term). Rewrote the comment to spell out the rule in plain English:

> Drop a headline if no real option row survived the filter under it (looking ahead until the next headline or divider). Buttons don't count — a section that ends up containing only an action button still reads as an orphaned header to the user, so we collapse it too.

The Vue counterpart at `components/Select/vue/filter-items.ts` doesn't carry the comment so no change there. Logic is unchanged in both.

emilmilanov:
@mabela416 @ryanjwilke — re: the Atlassian-style preview-next-to-code ask above. This is a docs-site infra change (component code stays the same), so I'd handle it in a separate PR off `main`. Drafted a plan below for review before I start.

---

# Plan — Atlassian-style per-tile demos for the docs site

## Goal
Replace each component's single demo block with a **stacked list of tiles**, where each tile shows the live component on top and its source code below. The framework tabs (Vue / React) drive **both** the source code and the live preview, so switching to React swaps the entire demo block.

## Layout

```
<h1>Select</h1>
<FrameworkTabs />
  <F only="vue">
    <DemoTilesList component="Select" framework="vue" />
    <VueReadMe />
  </F>
  <F only="react">
    <DemoTilesList component="Select" framework="react" />
    <ReactReadMe />
  </F>
<InstructionsContent />
```

Demo moves **below the tabs** so the tabs become the page-wide framework switch and the demo participates via the same `<F>` mechanism — no new toggle wiring or global event bus.

## File layout

```
docs/src/demos/Select/
├── tiles.ts                 ← shared manifest, framework-agnostic
├── vue/
│   ├── Default.vue
│   ├── AllContentTypes.vue
│   └── … (one file per tile)
└── react/
    ├── Default.tsx
    ├── AllContentTypes.tsx
    └── …
```

`tiles.ts` exports `[{ name, title }, …]` once; the page loops it twice, once per `<F>` wrapper, loading from the matching framework directory.

## Why per-tile files (vs. one big file with region markers)

1. **Copy-paste fidelity.** The code block shown under each tile *is* the file — no extraction, no transformation. Atlassian's "Copy code" trust comes from this.
2. **Each tile is a sandbox starter.** Drop-in for a future "Edit in CodeSandbox" link.

Tradeoff accepted: ~30 small files per component instead of one large file.

## New / changed components

- `docs/src/components/DemoTile.astro` — new. Props: `component`, `name`, `framework`, `title?`. Renders title row → live preview (framed in existing indigo-tinted surface) → code block (Shiki-highlighted, same as existing fenced code).
- `docs/src/components/DemoTilesList.astro` — new. Loops `tiles.ts` and renders one `<DemoTile>` per entry.
- `docs/src/components/Demo.vue` — extended glob to `../demos/*/vue/*.vue` and a new `name` prop alongside the existing `component` prop.
- `docs/src/components/DemoReact.tsx` — new. React parallel of `Demo.vue`, lazy-glob over `../demos/*/react/*.tsx`. Mounted via `client:only="react"` (already supported by Astro config).
- `docs/src/pages/components/[component].astro` — move the demo block below `<FrameworkTabs>`, wrap each in its `<F>`.

Source-code lookup for the code blocks:

```ts
const sources = {
  vue:   import.meta.glob('../demos/*/vue/*.vue',   { query: '?raw', import: 'default', eager: true }),
  react: import.meta.glob('../demos/*/react/*.tsx', { query: '?raw', import: 'default', eager: true }),
}
```

## Migration: Select

- Walk through current `docs/src/demos/Select.vue`. For every `<p class="…">SECTION NAME</p>` + adjacent `<Select>` block, extract into `docs/src/demos/Select/vue/<TileName>.vue` as a self-contained file (its own items array, handlers, refs — no shared module). Delete the original `Select.vue` after.
- React tiles ship incrementally in follow-up PRs. During the gap, `<DemoTile>` renders a "React demo coming — see Vue tab" placeholder for any tile missing under `react/`. Gaps tracked in `docs/src/demos/MISSING.md`.

## Other components

Out of scope. Their pages keep the current single-demo behavior. They migrate to `<DemoTile>` later, one PR per component.

## Scope of this PR

- New `<DemoTile>`, `<DemoTilesList>`, `<DemoReact>` components.
- Updated `[component].astro` page template (demo moves under tabs).
- Updated `Demo.vue` (extended glob).
- Select migration: Vue tiles split out fully; `tiles.ts` manifest written.
- React tiles: zero in this PR (placeholder shown). Follow-up PRs port them.
- No changes to Button, Tabs, etc.

components/Select/vue/_SelectOptionItem.vue
```diff
@@ -0,0 +1,234 @@
...
+    :id="id"
+    :role="isCustomSelectable ? 'option' : 'presentation'"
+    :aria-selected="isCustomSelectable ? selected : undefined"
+    :data-selected="selected || undefined"
+    :data-focused="focused || undefined"
+    :class="itemClasses"
+    @click="isCustomSelectable ? onClick($event) : undefined"
+    @mousedown="onMouseDown"
+  >
+    <component :is="(item.render as any)({ selected })" />
```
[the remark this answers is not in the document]
emilmilanov: Fixed. `<component :is>` was the wrong vehicle for arbitrary render output — replaced it with a minimal functional component (`CustomRender`) that just calls the consumer's `render` function inside `setup()` and returns the result. That way Vue normalizes the output the same way it does for any render function, so VNodes, strings, fragments, and `null` all render correctly.

```vue
const CustomRender = defineComponent({
  props: {
    render: { type: Function as ..., required: true },
    selected: { type: Boolean, required: true },
  },
  setup(props) {
    return () => props.render({ selected: props.selected })
  },
})
```

Template now reads `<CustomRender :render="item.render" :selected="selected" />`. The demo doesn't exercise `type: 'custom'` rows yet, so this was latent — verified the rest of the Vue Select rendering still works on the docs preview (all 8 listboxes render, no console errors).

components/Select/react/Select.tsx
mabela416: I prefer option 2
I don't like option 1 because with Omit and Pick, it is sometimes hard to figure out what props the type has so it's hard to follow.
Option 3 is fine but we'd need to omit fields again which has the same caveat as option 1
I'm okay with the tradeoff of spreading the docs across a few interfaces because it'd be easier to understand the types with the small named prop groups.

and just fyi this all personal preference so neither option is wrong

emilmilanov: Thanks @mabela416 — option 2 it is. Landed in [`549c89a0`](https://github.com/cypress-io/cypress-design/commit/549c89a0):

5 named groups in `@cypress-design/constants-select`:
- `SelectThemingProps` — `theme`, `size`
- `SelectHeaderProps` — `headerTitle/Button/IconLeft/Tag/IconRight`, `headerTabs/headerActiveTab`
- `SelectSearchProps` — `searchable`, `searchPlaceholder`, `searchFilters`
- `SelectFooterProps` — `footerLabel`, `footerAction`
- `SelectSizingProps` — `width/minWidth/maxWidth/height/maxHeight` (already existed; just doc-grouped with the new neighbors)

`SelectProps` and `SelectOptionListProps` now compose via `extends` (React) / intersection (Vue `defineProps<A & B & …>()`) instead of repeating fields. The icon types use the existing `IconNode` (= `unknown`) so the groups stay framework-agnostic; the React renderer narrows to `React.ComponentType` at the destructure boundary.

Net diff: 140 insertions / 172 deletions (more readable, no field listed in more than one place). Docs preview unchanged visually, 51/51 Cypress tests still green.

components/Select/vue/_SelectOptionItem.vue
```diff
@@ -0,0 +1,257 @@
...
+    <component
+      :is="item.iconRight"
+      v-if="item.iconRight"
+      size="16"
+      :interactive-colors-on-group="true"
+      :class="['ml-auto', iconColorClass]"
+    />
+    <span v-if="item.slotRight" class="ml-auto shrink-0">
+      <component :is="item.slotRight" />
+    </span>
```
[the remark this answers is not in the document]
emilmilanov: Fixed in [`30d5b5a0`](https://github.com/cypress-io/cypress-design/commit/30d5b5a0). Same pattern as the `item.render` (custom row) fix: wrap the consumer's `slotRight` in a small functional component (`NodeRender`) whose `setup()` returns `props.node` directly, so Vue accepts VNodes, strings, fragments, or anything else the React side does — not just component definitions.

components/Select/react/SelectOptionList.tsx
```diff
@@ -0,0 +1,343 @@
...
+          </div>
+        ) : (
+          filteredItems.map((item, index) => {
+            const itemValue = SelectConstants.getItemValue(item)
+            // Empty string is a valid `value` per `SelectItemDefault` —
+            // `itemValue === value` already excludes `undefined` ↔ '' (since
+            // `'' !== undefined`), so no extra guard is needed.
+            const selected = itemValue !== undefined && itemValue === value
+            const focused = index === focusedSelectableIndex
+            const rowId = itemIdPrefix ? `${itemIdPrefix}-${index}` : undefined
```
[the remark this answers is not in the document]
emilmilanov: Real bug — fixed in [`30d5b5a0`](https://github.com/cypress-io/cypress-design/commit/30d5b5a0). Lifted `searchValue` up to `Select` (was owned by `SelectOptionList`), so the keyboard handler and the rendered popover now work off the same filtered list:

- `Select` owns `searchValue`, computes `displayItems = filterAndCollapseHeadlines(items, searchValue)` once, walks `selectableIndices` on `displayItems`, and `Enter` resolves to `displayItems[realIndex]`.
- `SelectOptionList` becomes pure presentation — no internal search state, no internal filter. Receives `items` (already filtered), plus a controlled `searchValue` + `onSearchValueChange` (Vue: `@update:search-value`) for the search Textbox.

Applied symmetrically to Vue (`Select.vue` + `_SelectOptionList.vue`) and React (`Select.tsx` + `SelectOptionList.tsx`). Removed the "known limitation" comment from `Select.tsx` that flagged this bug.

Verified in the docs preview against the "With header + footer" demo: typing "with" filters down to one row, ArrowDown lands focus on that row, and `aria-activedescendant` matches the rendered DOM id.

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,395 @@
...
+        square={isTriggerIconOnly}
+        disabled={disabled}
+        type="button"
+        aria-haspopup="listbox"
+        aria-expanded={open}
+        aria-controls={popoverId}
+        aria-activedescendant={
+          open && focusedIndex >= 0
+            ? `${itemIdPrefix}-${selectableIndices[focusedIndex]}`
+            : undefined
```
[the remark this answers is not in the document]
emilmilanov: Fixed in [`aef6b40c`](https://github.com/cypress-io/cypress-design/commit/aef6b40c). Switched the reset dep from `searchValue` to `displayItems`, so the focus index now resets on **either** trigger — search-query change or consumer-swapped `items` (e.g., a header-tab change). Same change in Vue (`watch(displayItems, ...)`) and React (`useEffect(..., [displayItems])`).

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,395 @@
...
+      e.preventDefault()
+      setFocusedIndex((i) => {
+        const n = selectableIndices.length
+        if (n === 0) return -1
+        // First arrow press (i === -1) lands on the last option — wraps
+        // symmetrically with ArrowDown-from-last.
+        if (i === -1) return n - 1
+        return (i - 1 + n) % n
+      })
+      return
```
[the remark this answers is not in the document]
emilmilanov: This is intentional — for a popover with `searchable=true`, routing ArrowUp/Down from the search input to the listbox is the standard text-field + listbox pattern (Radix, Headless UI, native browser `<input list>` etc. all do this). A single-line `<input type="text">` has no useful behavior for vertical arrows anyway — they don't move the cursor by line — so the wrapper handler isn't stealing anything semantically meaningful. The instructions document this:

> Open with no row focused. The first `ArrowDown` lands focus on the first interactive row...

Leaving as-is. If we ever support a multi-line search field this would need revisiting.

components/Select/react/SelectOptionList.tsx
```diff
@@ -0,0 +1,338 @@
...
+            const key = (() => {
+              if ('key' in item && item.key) return item.key
+              if (itemValue) return itemValue
+              return `${item.type ?? 'default'}-${index}`
```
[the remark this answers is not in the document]
emilmilanov: Fixed in [`aef6b40c`](https://github.com/cypress-io/cypress-design/commit/aef6b40c). Both `itemKey` implementations (React `SelectOptionList.tsx` and Vue `_SelectOptionList.vue`) now use `itemValue !== undefined` instead of a truthy check, so a row with `value: ''` keeps its stable key based on the empty string rather than falling back to `${type}-${index}`.

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,397 @@
...
+    if (e.key === 'Enter') {
+      e.preventDefault()
+      // No-op when nothing is focused yet (user hasn't pressed an arrow).
+      if (focusedIndex === -1) return
+      const realIndex = selectableIndices[focusedIndex]
+      if (realIndex === undefined) return
+      const item = displayItems[realIndex]
+      handleSelect(item)
+    }
+  }
```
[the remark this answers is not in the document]
emilmilanov: Fixed in [`660eb994`](https://github.com/cypress-io/cypress-design/commit/660eb994). The open-state keyboard handler now matches `e.key === 'Enter' || e.key === ' '`. Both call `preventDefault` and run `handleSelect` on the focused row — Space no longer falls through to native button-activation on the trigger. Applied symmetrically in Vue (`Select.vue`) and React (`Select.tsx`).

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,397 @@
...
+  const toggle = () => setOpen(!open)
+  const close = () => setOpen(false)
+  const triggerCtx: SelectTriggerContext = { open, selected, toggle, close }
+
+  const defaultTriggerLabel =
+    selected && 'label' in selected ? selected.label : placeholder ?? ''
+  // When the trigger has no label text (no selection AND no placeholder),
+  // render it as an icon-only square so the chevron alone doesn't sit in a
+  // stretched-out button.
+  const isTriggerIconOnly = !defaultTriggerLabel
```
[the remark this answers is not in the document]
emilmilanov: Fixed in [`660eb994`](https://github.com/cypress-io/cypress-design/commit/660eb994). Added optional `label?: string` to `SelectItemCustom`; `getItemLabel` now returns it for custom rows, so `defaultTriggerLabel`'s `'label' in selected` check picks it up automatically. Consumers who want richer trigger content (icons, badges, etc.) can still use the `trigger` slot/render-prop. instructions.md documents the new field.

Trade-off: a custom row without `label` set still leaves the trigger showing the placeholder (same as before). That feels correct — we don't have a sensible string to derive from `render`'s arbitrary output. The `trigger` escape hatch is the right tool for fully-custom triggers.

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,400 @@
...
+      })
+      return
+    }
+    if (e.key === 'Enter' || e.key === ' ') {
+      // Both Enter and Space confirm the focused row. Without preventDefault,
+      // Space falls through to the native button-activation behavior on the
+      // focused trigger and toggles the popover closed without selecting.
+      e.preventDefault()
+      // No-op when nothing is focused yet (user hasn't pressed an arrow).
+      if (focusedIndex === -1) return
```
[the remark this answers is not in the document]
emilmilanov: Fixed in [`547b22e6`](https://github.com/cypress-io/cypress-design/commit/547b22e6). The previous round routed Space through the open-state handler but called `preventDefault` before the focus check — so spaces typed into the search input bubbled up and got swallowed. Moved `preventDefault` to AFTER the no-focus early return:

- focused row + Enter/Space → preventDefault + handleSelect ✓
- no row focused + Enter/Space → fall through (native behavior — spaces insert into the search input, Enter/Space on the trigger button toggles the popover) ✓

Same change in Vue (`Select.vue`) and React (`Select.tsx`). 59/59 Cypress tests still passing.

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,403 @@
...
+        return (i - 1 + n) % n
+      })
+      return
+    }
+    if (e.key === 'Enter' || e.key === ' ') {
+      // No-op when nothing is focused yet — and crucially, do NOT
+      // preventDefault. The wrapper onKeyDown also fires for keys typed
+      // INTO the header search input (events bubble), so swallowing
+      // Space here would block multi-word search queries.
+      if (focusedIndex === -1) return
```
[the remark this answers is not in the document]
emilmilanov: Fixed in [`cb5ed03f`](https://github.com/cypress-io/cypress-design/commit/cb5ed03f). The previous round swung the other way — dropped preventDefault entirely for the no-focus case to unblock spaces in search, but that let the trigger button's native activation close the panel. Now the no-focus branch is target-aware:

```ts
if (focusedIndex === -1) {
  const target = e.target as HTMLElement | null
  const isTextInput =
    target instanceof HTMLInputElement &&
    (target.type === 'text' || target.type === 'search')
  if (!isTextInput) e.preventDefault()
  return
}
```

- Trigger button → `preventDefault` so a second Enter/Space doesn't toggle closed ✓
- Search input → fall through, native space-insertion runs ✓

Verified both behaviors in the docs preview against the searchable "With header + footer" demo and the trigger-only "Pick one" demo. 59/59 Cypress tests still pass.

components/Select/react/SelectOptionList.tsx
```diff
@@ -0,0 +1,342 @@
...
+          )}
+        </div>
+      )}
+
+      <div className={SelectConstants.CssItemsContainerClasses}>
+        {items.length === 0 ? (
+          <div className={SelectConstants.CssEmptyStateClasses[theme]}>
+            No results
+          </div>
+        ) : (
```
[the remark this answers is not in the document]
emilmilanov: Fixed in [2382b5eb](https://github.com/cypress-io/cypress-design/commit/2382b5eb): predicate now keys off `selectableIndices.length === 0` and the items list renders unconditionally, so leftover divider/button rows stay visible alongside the "No results" hint. Verified in the docs preview.

components/Select/react/SelectOptionItem.tsx
```diff
@@ -0,0 +1,273 @@
...
+    // look puny in a `size=40` panel.
+    const buttonSize = size === '40' ? '32' : '24'
+    return (
+      <div
+        className={clsx(
+          SelectConstants.CssButtonRowClasses,
+          SelectConstants.CssOptionItemHeightClasses[size],
+          SelectConstants.CssOptionItemPaddingClasses[size],
+        )}
+      >
```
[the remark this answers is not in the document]
emilmilanov: Fixed in [37ad6028](https://github.com/cypress-io/cypress-design/commit/37ad6028): button-row wrapper now sets `role="presentation"` in both Vue and React, matching headline/divider so the listbox tree only exposes selectable rows to AT.

components/Select/constants/src/index.ts
```diff
@@ -0,0 +1,251 @@
...
+export type SelectSize = keyof typeof CssOptionItemHeightClasses // '32' | '40'
+export type SelectAlignment = 'left' | 'right'
+
+// `IconNode` accepts any consumer-renderable node. Constants are framework-
+// agnostic, so the type is intentionally loose — both React (functional
+// components, JSX nodes) and Vue (`DefineSetupFnComponent`-flavored shapes
+// from `@cypress-design/vue-icon`) need to fit through. Each framework's
+// renderer does its own runtime dispatch (`<component :is>` in Vue,
+// `React.createElement` in React), so the type is just a passthrough.
+export type IconNode = unknown
```
mabela416, line 22: Not the biggest fan of these loose typings but I understand it's to maintain the shared types between Vue and React. There are other options on how to get the correct typings for both frameworks but it would require creating more types and I don't think it's worth it for this component because of how nested some of types are and how many use the same loose types so I think it would get hard to read. Just bringing it up to raise awareness but for components that have simpler types (no nesting or not too much sharing), we should avoid this

components/Select/constants/src/index.ts
```diff
@@ -0,0 +1,251 @@
...
+  | SelectItemDefault
+  | SelectItemCheckbox
+  | SelectItemUser
+  | (SelectItemCustom & { value: string })
+
+// Helpers used by both React and Vue implementations to filter or classify items.
+export function isSelectable(item: SelectItem): item is SelectableItem {
+  if (item.type === 'default' || item.type === undefined) return !item.disabled
+  if (item.type === 'checkbox') return !item.disabled
+  if (item.type === 'user') return true
```
mabela416: why doesn't the user type also check for item.disabled?

components/Select/constants/src/index.ts
```diff
@@ -0,0 +1,251 @@
...
+ * `undefined` so `buildSizingStyle` omits the property entirely instead of
+ * shipping `'NaNpx'` or `''` (browsers ignore both, but a missing key is
+ * cleaner than a silently dropped one).
+ */
+export function toCssLength(value: CssLength | undefined): string | undefined {
+  if (value === undefined) return undefined
+  if (typeof value === 'number') {
+    return Number.isFinite(value) ? `${value}px` : undefined
+  }
+  return value === '' ? undefined : value
```
mabela416: just a nit. in line 218, we can both check `if(value === undefined || value === '') return undefined`  and in line 222 just do `return value`

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,414 @@
...
+  className?: string
+  popoverClassName?: string
+  id?: string
+}
+
+let uidCounter = 0
+const useUid = () => {
+  const [uid] = React.useState(() => `cy-select-${++uidCounter}`)
+  return uid
+}
```
mabela416: React introduced a hook to do this https://react.dev/reference/react/useId so we should use that instead

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,414 @@
...
+        aria-expanded={open}
+        aria-controls={popoverId}
+        aria-activedescendant={
+          open && focusedIndex >= 0
+            ? `${itemIdPrefix}-${selectableIndices[focusedIndex]}`
+            : undefined
+        }
+        onClick={toggle}
+        className={
+          isTriggerIconOnly ? undefined : SelectConstants.CssTriggerWidthClasses
```
mabela416: What do you think about also giving the client the option to pass a classname for the trigger?

components/Select/react/SelectOptionItem.tsx
```diff
@@ -0,0 +1,277 @@
...
+      <div
+        role="presentation"
+        className={clsx(
+          SelectConstants.CssButtonRowClasses,
+          SelectConstants.CssOptionItemHeightClasses[size],
+          SelectConstants.CssOptionItemPaddingClasses[size],
+        )}
+      >
+        <Button
+          variant={buttonVariant as never}
```
mabela416: You can import the `ButtonVariants` and do ` as ButtonVariants` instead

components/Select/react/SelectOptionItem.tsx
```diff
@@ -0,0 +1,277 @@
...
+        >
+          {(() => {
+            // Render iconLeft via PascalCase JSX so the icon can be either a
+            // component reference (the design-system convention) or any other
+            // ReactNode the consumer chooses to pass — matches the header-row
+            // icon rendering. Avoids `React.createElement` blowing up when
+            // someone hands us `<MyIcon />` (a JSX element) instead of
+            // `MyIcon` (the component).
+            const IconLeft = item.iconLeft as
+              | React.ComponentType<Record<string, unknown>>
```
mabela416: Instead of this type, we can do this to get more accurate typing here


```
import type { OpenIconProps } from '@cypress-design/icon-registry'

const IconLeft = item.iconLeft as
  | React.ComponentType<Omit<OpenIconProps, 'name'>>
  | undefined
```


OR you could just define the props that this will need

```
React.ComponentType<{ size?: string; interactiveColorsOnGroup?: boolean;  }>
```

We want to avoid these `unknown` or `any` types whenever possible. It's easier to catch issues at compile time when the types are defined. Also intellisense and autocomplete works better when types are well defined so it makes it easier for the client to know exactly what needs to be passed without having to check the docs.

components/Select/react/SelectOptionList.tsx
```diff
@@ -0,0 +1,345 @@
...
+  // Keyboard-focused row index (within the rendered list).
+  // Provided by Select; the panel just renders the state.
+  focusedIndex?: number
+  itemIdPrefix?: string
+}
+
+// React JSX needs a component type to instantiate; the shared
+// SelectHeaderProps interface uses IconNode (= unknown) so it stays
+// framework-agnostic. Re-narrow at the destructure boundary.
+type IconComponent = React.ComponentType<Record<string, unknown>>
```
mabela416: Same comment here, we could do 
```
type IconComponent = React.ComponentType<
  Omit<OpenIconProps & { className?: string }, 'name'>>
>
```


I can also foresee us having to create this "IconComponent" type for other future components so it might be worth to create this type in the `icon-registry` package and then just import it from there for both `SelectOptionList` and `SelectOptionItem`

components/Select/vue/_SelectOptionItem.vue
```diff
@@ -0,0 +1,271 @@
...
+// Vue's `<component :is>` expects a component definition, not arbitrary
+// render output (VNodes, strings, fragments). Wrapping the consumer's
+// `render` function in a minimal functional component lets us render
+// whatever it returns — h() output, strings, fragments, or null.
+const CustomRender = defineComponent({
+  props: {
+    render: {
+      type: Function as unknown as () => (ctx: {
+        selected: boolean
+      }) => unknown,
```
mabela416: Instead of casting this as `unknown`, you could do 
```
 type: Function as PropType<(ctx: { selected: boolean }) => unknown>,
```

components/Select/vue/_SelectOptionItem.vue
```diff
@@ -0,0 +1,271 @@
...
+  setup(props) {
+    return () => props.render({ selected: props.selected })
+  },
+})
+
+// Same trick for `slotRight` on default rows. `<component :is>` only works
+// when the value is a component definition; with this helper Vue happily
+// renders VNodes, strings, fragments, or anything the consumer hands in.
+const NodeRender = defineComponent({
+  props: { node: { required: true, type: null as unknown as () => unknown } },
```
mabela416: This `type` prop is not being used anywhere so we can drop it
```
props: { node: { required: true } }
```

components/Select/vue/_SelectOptionItem.vue
```diff
@@ -0,0 +1,271 @@
...
+    role="presentation"
+    :class="[
+      SelectConstants.CssButtonRowClasses,
+      SelectConstants.CssOptionItemHeightClasses[size],
+      SelectConstants.CssOptionItemPaddingClasses[size],
+    ]"
+  >
+    <Button
+      :variant="
+        (item.variant as never) ?? (theme === 'dark' ? 'outline-dark' : 'white')
```
mabela416: Same comment as the React component. We can do 
```
as ButtonVariants
```

components/Select/vue/_SelectOptionList.vue
```diff
@@ -0,0 +1,291 @@
...
+      <div
+        v-if="selectableIndices.length === 0"
+        :class="SelectConstants.CssEmptyStateClasses[theme]"
+      >
+        No results
+      </div>
+      <SelectOptionItem
+        v-for="(item, index) in items"
+        :key="itemKey(item, index)"
+        :id="rowId(index)"
```
mabela416: can you fix the ordering here so that we don't have this warning

components/Select/react/SelectOptionList.tsx
```diff
@@ -0,0 +1,345 @@
...
+                <div className="self-start">
+                  <Tabs
+                    variant={
+                      theme === 'dark'
+                        ? size === '40'
+                          ? 'dark-large'
+                          : 'dark-small'
+                        : 'default'
+                    }
+                    tabs={headerTabs as never}
```
mabela416: We can use 
```
import type { Tab } from '@cypress-design/constants-tabs'

as Tab[]
```

components/Select/vue/_SelectOptionList.vue
```diff
@@ -0,0 +1,291 @@
...
+        <div v-if="headerTabs && headerTabs.length > 0" class="self-start">
+          <Tabs
+            :variant="
+              theme === 'dark'
+                ? size === '40'
+                  ? 'dark-large'
+                  : 'dark-small'
+                : 'default'
+            "
+            :tabs="headerTabs as never"
```
mabela416: Same comment as the react componet.

We can use 
```
import type { Tab } from '@cypress-design/constants-tabs'

as Tab[]
```

components/Select/vue/_SelectOptionList.vue
```diff
@@ -0,0 +1,291 @@
...
+})
+
+const emit = defineEmits<{
+  (e: 'select', item: SelectItem): void
+  (e: 'header-tab-change', id: string): void
+  (e: 'update:searchValue', value: string): void
+}>()
+
+defineSlots<{
+  footer?: () => unknown
```
mabela416: this can be `footer?: () => VNode[]`

components/Select/vue/filter-items.ts
mabela416, line 1: this file has the exact same logic as the one for react. I think we could add this function to the `constants` package so that React and Vue can both use it and we can have this function in one place instead of two duplicate copies.

components/Select/vue/Select.vue
```diff
@@ -0,0 +1,398 @@
...
+    <div ref="triggerRef">
+      <slot
+        name="trigger"
+        :open="open"
+        :selected="selected"
+        :toggle="toggle"
+        :close="close"
+      >
+        <Button
+          :variant="triggerVariant as never"
```
mabela416: We can use `ButtonVariants` here as well

components/Select/vue/Select.vue
```diff
@@ -0,0 +1,398 @@
...
+
+// ---------- ids ----------
+let uidCounter = 0
+const uid = `cy-select-${++uidCounter}`
+const popoverId = computed(() =>
+  props.id ? `${props.id}-popover` : `${uid}-popover`,
+)
+const itemIdPrefix = computed(() =>
+  props.id ? `${props.id}-opt` : `${uid}-opt`,
+)
```
mabela416: Vue also has a `useId` function https://skstory.online/p/understanding-useid-in-vue3/

components/Select/vue/Select.vue
```diff
@@ -0,0 +1,398 @@
...
+  align: SelectConstants.DefaultAlignment,
+  triggerVariant: SelectConstants.DefaultTriggerVariant,
+  searchable: false,
+  // Vue 3 quirk: Boolean props bound via `v-bind` with `undefined` are
+  // cast to `false` BEFORE the child's withDefaults applies. Without an
+  // explicit default here, omitting `searchFilters` at the call site
+  // means Select forwards `false` to SelectOptionList, disabling the
+  // filter (the row stays visual-only). Default true so search actually
+  // filters by default.
+  searchFilters: true,
```
mabela416, line 63: When searchFilters isn't passed, Vue coerces the optional boolean prop to false, making that expression false even when searchable is true. But _SelectOptionList already Omits searchFilters — it never receives it — so the comment's claim about "forwarding" is stale.

So we don't need to do this if we do this in line [151](https://github.com/cypress-io/cypress-design/pull/668/changes#diff-0f03742a89572c2377d88828d2855329622fed5fae4853212a49ed611afb3326R158)

`props.searchable && props.searchFilters !== false`

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,414 @@
...
+  }, [open])
+
+  // Anything that shrinks the visible list — typing into search or the
+  // consumer swapping `items` (e.g., a header-tab change) — can leave
+  // `focusedIndex` past the end of `selectableIndices`, which makes
+  // `aria-activedescendant` resolve to an id ending in `undefined` and
+  // hides the focus ring. Reset to -1 whenever the displayed list changes.
+  React.useEffect(() => {
+    setFocusedIndex(-1)
+  }, [displayItems])
```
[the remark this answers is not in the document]
emilmilanov: Fixed in [072595f3](https://github.com/cypress-io/cypress-design/commit/072595f3): both React and Vue now derive a stable content signature (`type|key/value|label` per row) and key the reset effect on that. Inline-array re-renders with identical rows no longer drop focus or `aria-activedescendant`; real changes (search filter, header-tab swap) still reset.

components/Select/constants/src/index.ts
emilmilanov: Acknowledged — we're deliberately living with the `unknown`/loose typings here to keep `SelectItem` framework-agnostic and shared across React + Vue without forcing every consumer through framework-specific generics. Mabel's call: for simpler components we should reach for the tighter typing instead. Following that rule going forward.

components/Select/constants/src/index.ts
emilmilanov: Fixed in [84336cc5](https://github.com/cypress-io/cypress-design/commit/84336cc5): added `disabled?: boolean` to `SelectItemUser`, updated `isSelectable` to honor it, and both `SelectOptionItem` implementations now apply the disabled state + `aria-disabled` + click short-circuit consistent with default/checkbox rows.

components/Select/constants/src/index.ts
emilmilanov: Fixed in [84336cc5](https://github.com/cypress-io/cypress-design/commit/84336cc5): collapsed the two cases into a single early return — `if (value === undefined || value === '') return undefined` and the trailing branch is now just `return value`.

components/Select/react/Select.tsx
emilmilanov: Fixed in [8192a357](https://github.com/cypress-io/cypress-design/commit/8192a357): React Select now uses `React.useId()` and namespaces it as `cy-select-<id>` to keep the DOM ids readable.

components/Select/react/Select.tsx
emilmilanov: Done in [8192a357](https://github.com/cypress-io/cypress-design/commit/8192a357): added `triggerClassName?: string` (React) — applied via `clsx` so it composes with the default trigger width classes. Lets consumers tweak the default Button trigger without taking over the whole `trigger` slot.

components/Select/react/SelectOptionItem.tsx
emilmilanov: Fixed in [8192a357](https://github.com/cypress-io/cypress-design/commit/8192a357): now `as ButtonVariants` (imported from `@cypress-design/constants-button`).

components/Select/react/SelectOptionItem.tsx
emilmilanov: Fixed in [8192a357](https://github.com/cypress-io/cypress-design/commit/8192a357): pulled `IconComponent` up to a module-local type alias `React.ComponentType<Omit<OpenIconProps, 'name'>>` (from `@cypress-design/icon-registry`). The narrow `Record<string, unknown>` cast on `Icon` is replaced with this, so intellisense for `size` / `interactiveColorsOnGroup` works in both the iconLeft JSX and the `renderIcon` helper.

components/Select/react/SelectOptionList.tsx
emilmilanov: Fixed in [8192a357](https://github.com/cypress-io/cypress-design/commit/8192a357): React `SelectOptionList` now uses `React.ComponentType<Omit<OpenIconProps, 'name'> & { className?: string }>` for `IconComponent`, narrowing all three header icon slots. Re: lifting `IconComponent` into `icon-registry` — agree it's the right place long-term; deferring to a follow-up PR so we don't expand this PR further.

components/Select/vue/_SelectOptionItem.vue
emilmilanov: Fixed in [8192a357](https://github.com/cypress-io/cypress-design/commit/8192a357): switched the Vue `CustomRender` helper to `Function as PropType<(ctx: { selected: boolean }) => unknown>`.

components/Select/vue/_SelectOptionItem.vue
emilmilanov: Fixed in [8192a357](https://github.com/cypress-io/cypress-design/commit/8192a357): dropped the unused `type` field on `NodeRender`'s `node` prop.

components/Select/vue/_SelectOptionItem.vue
emilmilanov: Fixed in [8192a357](https://github.com/cypress-io/cypress-design/commit/8192a357): now `as ButtonVariants` (imported from `@cypress-design/constants-button`).

components/Select/vue/_SelectOptionList.vue
emilmilanov: Fixed in [8192a357](https://github.com/cypress-io/cypress-design/commit/8192a357): `:id` now sits before `:key` on the v-for'd `<SelectOptionItem>`. `eslint components/Select/vue/_SelectOptionList.vue` is now clean.

components/Select/react/SelectOptionList.tsx
emilmilanov: Fixed in [8192a357](https://github.com/cypress-io/cypress-design/commit/8192a357): React side imports `Tab` from `@cypress-design/constants-tabs` and casts `headerTabs as Tab[]`.

components/Select/vue/_SelectOptionList.vue
emilmilanov: Fixed in [8192a357](https://github.com/cypress-io/cypress-design/commit/8192a357): same `Tab[]` cast on the Vue side via `import type { Tab } from '@cypress-design/constants-tabs'`.

components/Select/vue/_SelectOptionList.vue
emilmilanov: Fixed in [8192a357](https://github.com/cypress-io/cypress-design/commit/8192a357): footer slot type is now `() => VNode[]` (`type VNode` imported from `vue`).

components/Select/vue/filter-items.ts
emilmilanov: Done in [84336cc5](https://github.com/cypress-io/cypress-design/commit/84336cc5): `filterAndCollapseHeadlines` + `getSelectableIndices` now live in `@cypress-design/constants-select` (with `getInteractiveIndices` joining them on the accessibility branch). Both duplicate `react/filter-items.ts` and `vue/filter-items.ts` are deleted; Select and SelectOptionList in both frameworks import directly from the constants package.

components/Select/vue/Select.vue
emilmilanov: Fixed in [8192a357](https://github.com/cypress-io/cypress-design/commit/8192a357): Vue Select.vue now casts `triggerVariant as ButtonVariants` via the typed import from `@cypress-design/constants-button`.

components/Select/vue/Select.vue
emilmilanov: Vue 3.5 introduces `useId`, but cypress-design pins Vue 3.4 — `useId` is undefined there and breaks at runtime. I used `getCurrentInstance()?.uid` instead, which is what Vue's own `useId` is built on and works in 3.x. Same per-instance, monotonically-increasing id; namespaced as `cy-select-<n>`. Happy to swap to `useId` whenever the Vue 3.5 bump lands.

components/Select/vue/Select.vue
emilmilanov: Fixed in [8192a357](https://github.com/cypress-io/cypress-design/commit/8192a357): comment is now accurate. `_SelectOptionList` does Omit `searchFilters` and never receives it — the default lives on `Select.vue` because Vue 3 coerces an unset Boolean prop to `false`, so we need an explicit `searchFilters: true` to keep the filter on by default. Note: I tried Mabel's suggested `searchFilters !== false` without the default and verified it breaks default filtering (omitted prop → `false` → expression is false → filter off), so reverted and kept the default with a clearer comment.

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,430 @@
...
+      }
+      const realIndex = selectableIndices[focusedIndex]
+      if (realIndex === undefined) return
+      // A row is focused. Both Enter and Space confirm. preventDefault
+      // suppresses the native button-activation that would otherwise toggle
+      // the popover closed without selecting.
+      e.preventDefault()
+      const item = displayItems[realIndex]
+      handleSelect(item)
+    }
```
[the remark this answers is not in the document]
emilmilanov: Fixed in [6985137d](https://github.com/cypress-io/cypress-design/commit/6985137d): lifted the text-input check above the focused-index branch in both React and Vue. Bubbled Enter/Space from the header `Textbox` now returns early regardless of whether a row is focused, so the search input keeps its character (Space) or default submit (Enter) and the highlighted option is no longer hijacked.

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,433 @@
...
+    // would still see the cleared trigger fall back to the last clicked
+    // row — exactly the "stale label" bug Cursor caught).
+    if (!isValueControlled) setInternalValue(nextValue)
+    onChange?.(nextValue, item)
+    if (item.type !== 'checkbox') setOpen(false)
+  }
+
+  // ---------- Trigger rendering ----------
+  const toggle = () => setOpen(!open)
+  const close = () => setOpen(false)
```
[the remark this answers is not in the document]
emilmilanov: Fixed in [78d10f3c](https://github.com/cypress-io/cypress-design/commit/78d10f3c): `toggle()` in the trigger context now short-circuits when `disabled` is true (both React and Vue). `close()` stays unconditional — a parent should still be able to force-close an open popover after the Select is disabled mid-flight. Verified in preview: mutating `disabled` to `true` on a live custom-trigger Select, clicking the slotted button → popover doesn't open; restoring `disabled=false` → toggle works again.

mabela416 (approved):

components/Select/vue/Select.vue
```diff
@@ -0,0 +1,433 @@
...
+  props.id ? `${props.id}-opt` : `${uid}-opt`,
+)
+
+// ---------- open state ----------
+// Note: Vue normalizes absent Boolean props to `false`, so we can't reliably
+// distinguish "consumer didn't pass `open`" from "consumer passed `open=false`".
+// We use internal state as the single source of truth: `open` and `defaultOpen`
+// both seed the initial value, and updates flow out via `update:open`. A
+// consumer that wants to force-close the popover can re-key the component.
+const internalOpen = ref(props.open || props.defaultOpen)
```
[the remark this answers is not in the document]
emilmilanov: Fixed in [e46b4b24](https://github.com/cypress-io/cypress-design/commit/e46b4b24): switched the seed from `props.open || props.defaultOpen` to `props.open ?? props.defaultOpen` and added `open: undefined` to `withDefaults`. The explicit `undefined` default opts the prop out of Vue 3's Boolean coercion (which would otherwise map an absent prop to `false`), so `??` can distinguish `:open="false"` from an omitted prop. Verified at runtime: `typeof props.open === 'undefined'` on Select instances that don't pass it, so the fallback to `defaultOpen` only fires when `open` was actually omitted.

emilmilanov:
@mabela416 — thanks for the thorough review! Everything's addressed:

**Typing tightening** (replaced `as never` casts):
- `ButtonVariants` (from `@cypress-design/constants-button`) on `triggerVariant` and button-row `variant` across React + Vue ([8192a357](https://github.com/cypress-io/cypress-design/commit/8192a357))
- `Tab[]` (from `@cypress-design/constants-tabs`) on header tabs in both `SelectOptionList` files
- `IconComponent` typed as `React.ComponentType<Omit<OpenIconProps, 'name'> & { className?: string }>` in `SelectOptionList.tsx` + `SelectOptionItem.tsx` (with a note for follow-up: lifting `IconComponent` into `icon-registry`)
- Vue `CustomRender` uses `PropType<>` instead of `Function as unknown as …`
- Vue `NodeRender` drops the unused `type` field
- Footer slot typed as `() => VNode[]`

**Framework idiomatic ids**: React → `React.useId()`; Vue → `getCurrentInstance()?.uid` (Vue 3.4 doesn't ship `useId` yet — happy to swap when we bump to 3.5+)

**API**: added `triggerClassName?: string` so consumers can style the default trigger without taking the full `trigger` slot.

**Consolidation**: moved `filterAndCollapseHeadlines` + `getSelectableIndices` into `@cypress-design/constants-select` and deleted the duplicate `react/filter-items.ts` / `vue/filter-items.ts` ([84336cc5](https://github.com/cypress-io/cypress-design/commit/84336cc5)). Same commit adds `disabled?: boolean` to `SelectItemUser` (honored in `isSelectable` + click guards + `aria-disabled`) and collapses the `toCssLength` empty-string check.

**Plus the small ones**: `vue/attributes-order` fixed on the v-for'd `<SelectOptionItem>`; stale `searchFilters`-forwarding comment rewritten on `Select.vue`.

Plus a few from Cursor along the way (focus drop on inline-array re-renders, custom-trigger `toggle()` honoring `disabled`, Vue `open=false` honoring explicit binding, Enter vs Space split on search input). All resolved.

Ready for a re-review when you have time — no rush.

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,442 @@
...
+    const nextValue = isCheckboxToggle ? undefined : itemValue
+    // Uncontrolled consumers: track the selection internally. Controlled
+    // consumers: the parent's prop drives the next render; don't touch
+    // internalValue (otherwise a parent that clears `value` to `undefined`
+    // would still see the cleared trigger fall back to the last clicked
+    // row — exactly the "stale label" bug Cursor caught).
+    if (!isValueControlled) setInternalValue(nextValue)
+    onChange?.(nextValue, item)
+    if (item.type !== 'checkbox') setOpen(false)
+  }
```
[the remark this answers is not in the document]
emilmilanov: Fixed in [a64d9635](https://github.com/cypress-io/cypress-design/commit/a64d9635) — belt + suspenders in both React and Vue:

- A `useEffect` / `watch` on `disabled` closes the popover when it flips to true. The trigger and `toggle()` already short-circuit, but the already-mounted popover stayed visible — now it auto-closes, which matches what consumers expect from a disabled component and incidentally fixes the sibling Escape/Tab finding (no popover left to dismiss).
- `handleSelect` short-circuits on `disabled` so a click already in flight when the prop flipped still can't mutate the value.

Verified in preview by mutating `disabled` on a live "Pick one" Select: popover closes on flip, the disabled trigger can't reopen, re-enabling doesn't auto-reopen but the next click does.

components/Select/react/Select.tsx
```diff
@@ -0,0 +1,442 @@
...
+        triggerRef.current?.querySelector(
+          'button, [tabindex]',
+        ) as HTMLElement | null
+      )?.focus()
+      return
+    }
+    if (e.key === 'Tab') {
+      setOpen(false)
+      return
+    }
```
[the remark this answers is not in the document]
emilmilanov: Subsumed by the close-on-disabled-flip effect in [a64d9635](https://github.com/cypress-io/cypress-design/commit/a64d9635). With that effect, the popover auto-closes when `disabled` flips to true — so by the time the user reaches for Escape / Tab there's nothing to dismiss. Same root cause as the "Disabled select still accepts clicks" finding (PR #668 [discussion_r3499932934](https://github.com/cypress-io/cypress-design/pull/668#discussion_r3499932934)); the one fix covers both.

emilmilanov:
Superseded by [#697](https://github.com/cypress-io/cypress-design/pull/697) — the four staged PRs (#667, #668, #691, #692) have been consolidated into a single integration branch that will merge to `main`. The full review history from this PR is preserved and linked from #697. Closing here.

PR #672, opened 2026-05-29
docs(principles): turn feedback into reusable rules

ryanjwilke:
## What & why

I keep giving the same feedback in Slack, Zoom, Claude, and GitHub. The goal of this repo is to capture those points once so any agent reviewing Cypress work brings the same perspective I would — as close to cloning my judgment as the system allows.

For that to work, agents need a file that fires on _every_ feedback exchange, not only when someone says the word "principle." That's what `.agents/principles/learning-from-feedback.md` does. After reading it, a reviewer (human or agent) should:

- Treat every piece of feedback as a candidate rule, not just a task.
- Extract the underlying rule from the specific case.
- Test whether it would apply on another surface.
- If yes, vet it against the existing principles and open a PR to add it.

## How

- New file at `.agents/principles/learning-from-feedback.md` with a **Notice → Extract → Test → Codify** loop at the top, then the vetting checks (sounds-principled-but-isn't / conflicts / additive refinement) and the PR-as-record path for adding a new principle.
- Linked from `.agents/index.md` so the router actually surfaces it.
- Cross-linked from `principles/feedback.md` so anyone fetched for "giving feedback" also gets this.

Replaces what was removed from [cypress-io/cypress.io#2517](https://github.com/cypress-io/cypress.io/pull/2517)'s `design-system-access.md` (now correctly access-only). Complements #670 (which added "Apply the principle; don't cite it" to `feedback.md`).

## Where to focus review

- Is the trigger wide enough? Will an agent reaching this file act on the full Notice → Extract → Test → Codify loop, or stop at "no one called it a principle, skip"?
- Does the filename + index description make this discoverable from a cold start?
- Is anything here a restatement of something already in `feedback.md` or `ux.md` that should be removed?

---

> [!NOTE]
> **Low Risk**
> Changes are limited to agent guidance and GitHub CLI workflow docs; no application runtime, auth, or data paths are modified.
> 
> **Overview**
> Adds **`.agents/principles/learning-from-feedback.md`**, which tells agents to run **Notice → Extract → Test → Codify** on every feedback exchange (any channel, any author), vet candidates against existing principles, and record new rules via PR to `cypress-design`. The agent **index** and **`feedback.md`** now route to it; **`feedback.md`** clarifies when to quote principles vs apply them without citation.
> 
> In the same pass, **`cy-review-pr-comments`** is broadened from Copilot-only to **all unresolved review authors** (minus CI noise bots), with safer **paginated GraphQL**, **root-comment-only** threading, and a **`TOTAL_INSCOPE`** sanity check. Triage adds **human-high-signal** rules, mandatory **Step 4.5** user sign-off before replies/resolves, **reply-then-resolve** per thread, optional **Copilot re-review** via `gh pr edit`, and **hard stops** to avoid endless review loops.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2881877d853cf70c32908e85d5f35845552d7d58. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.agents/principles/feedback.md
```diff
@@ -23,6 +23,7 @@ description: Principles for reviewing other people's work and giving design / pr
 
 ## Related
 
+- [learning-from-feedback.md](./learning-from-feedback.md) — when feedback contains a reusable rule, capture it as a principle
```
ryanjwilke: Feedback from who? Me? Brian? Emil? Jennifer? Claude/Copilot?

.agents/principles/learning-from-feedback.md
```diff
@@ -0,0 +1,46 @@
...
+# Learning from feedback
+
+Most feedback is a one-off fix. Some of it is a rule in disguise — the same point you'd make again next week on a different surface. This file is about catching the second kind and writing it down, so the next reviewer (human or agent) starts from the same place.
+
+The principles are a living document. They grow from real feedback and decisions, not from theory.
+
+## When feedback arrives, run these four steps
+
+1. **Notice** — treat every piece of feedback as a candidate for a rule, not just a task to finish. This applies whether the feedback is framed as a principle or not. Most of it won't be.
+2. **Extract** — pull the underlying rule out of the specific case. "Move this label up" is a fix. "Labels sit above their input unless space forbids it" is a rule.
```
ryanjwilke: Move this label up is a fix, but there's a *why* behind it that could be stated by the human, asked by the agent to turn it into a rule.

.agents/principles/learning-from-feedback.md
```diff
@@ -0,0 +1,46 @@
...
+
+Before adding anything new, check it against the principles already exposed. Surface mismatches instead of going along quietly.
+
+- **Sounds principled but isn't there.** If something is asserted as a design / UX / product principle but isn't in the principle or pillar files, say so plainly: it's reasonable, but it's a _new_ rule, not an existing one. Don't dress a preference up as an existing principle, and don't invent a principle to justify a take.
+- **Conflicts with an existing principle.** If guidance contradicts a principle that _is_ exposed, name the specific principle and the tension — reasoning before verdict (see [feedback.md](./feedback.md)). The existing principle isn't automatically right; the point is to make the conflict explicit so it's resolved deliberately, not by accident.
+- **Additive refinement.** If a take is consistent but a principle would push it further, offer the refinement as additive — not a correction.
+
+Quote the principle verbatim in this internal check, so the source is verifiable. That's the one place to quote it — see **"Apply the principle; don't cite it"** in [feedback.md](./feedback.md) for why audience-facing feedback shouldn't.
```
ryanjwilke: Well, we want to quote the principle to see that the human agrees that we should document this as a new principle. (If that's what this is referring to). This is more something that could be used during reviews, replies to comments (by an agent), etc. Take feedback from human's as a moment to learn.

.agents/index.md
```diff
@@ -15,6 +15,7 @@ Foundational thinking that frames how we make design and product decisions. Fetc
 - [principles/ux.md](./principles/ux.md) — UX, product design, and business-goal principles (where business goals meet user needs, JTBD, restraint, pricing)
 - [principles/visual-hierarchy.md](./principles/visual-hierarchy.md) — directing the eye with color, icons, spacing, size
 - [principles/feedback.md](./principles/feedback.md) — reviewing work and giving feedback
+- [principles/learning-from-feedback.md](./principles/learning-from-feedback.md) — turning feedback into reusable rules; how to vet and add a new principle
```
ryanjwilke: We may need to find a better trigger for an agent to go and review this file.

.agents/principles/feedback.md
ryanjwilke: Fixed in 2881877d — added a line near the top of learning-from-feedback.md naming the sources (teammates by name, other agents, Copilot, automation), and tightened the link description here in feedback.md to say the same. Source is now explicit on both sides.

.agents/principles/learning-from-feedback.md
ryanjwilke: Fixed in 2881877d — rewrote Extract so the agent asks the human for the *why* behind the fix and turns that answer into the rule (instead of inferring it silently). The example now reads as a back-and-forth: "Move this label up" → agent asks "is the rule labels-sit-above-input, or one-off?" → human's answer is the rule.

.agents/principles/learning-from-feedback.md
ryanjwilke: Fixed in 2881877d — reframed the verbatim quoting. It's now a live, audience-facing confirmation step during the review or reply: the agent quotes the proposed/existing principle back to the human so they can confirm whether to document it. Dropped the "internal check" framing. Cross-reference to "Apply the principle; don't cite it" kept — that rule is for *applying* settled principles, while this quote is the moment a new one is being settled.

.agents/index.md
ryanjwilke: Fixed in 2881877d — strengthened the trigger in three places: the index bullet now reads "**fetch on every feedback exchange** (PR review, reply to a comment, Slack thread, Zoom recap)", the file's frontmatter description was rewritten the same way, and the cross-link in feedback.md picked up the same phrasing. An agent reaching any of these three entry points should now pull the file on any feedback exchange, not only when "principle" is named.

PR #687, opened 2026-06-15
feat(icons): add action-test-replay at 12px and 24px

ryanjwilke:
## What & why

The **Test Replay** icon (`action-test-replay`) only shipped at **16px** in the design system. Because 12px and 24px already exist in Figma but weren't published, [cypress.io](https://github.com/cypress-io/cypress.io) had to carry a local stand-in component (`src/components/product-shared/TestReplayIcon.vue`) with the SVGs inlined — its own header comment calls it temporary "until those variants exist upstream."

This adds **12px** and **24px** to the registry so consumers can use `IconActionTestReplay` with `size="12"` / `size="24"` directly and drop the stand-in.

## How

- Added `icon-registry/icons-static/action-test-replay_x12.svg` and `_x24.svg`. The build (`build-icons.mjs`) picks these up by filename and regenerates `icons.ts` — `availableSizes` is now `['12','16','24']`.
- **Geometry is the Figma export** (taken from the cypress.io stand-in, which had it inlined). The stand-in hardcoded hex (`#5A5F7A` / `#D0D2E0`); I converted both to the registry convention — `currentColor` + `class="icon-dark"` — so theming flows through `@cypress-design/css`.
- **24px** is a 2px-stroke line icon (filled play triangle + stroked circular arrow), matching the existing 16px construction. **12px** is a solid glyph (no stroke), consistent with other `_x12` solid icons and the right pattern at that size.
- **Dropped the 24px gray disc** the stand-in carried. That disc was optional/transparent-by-default there; the existing 16px has none. Single-tone across all three sizes matches the existing icon and keeps the generated type clean: `size?: '12' | '16' | '24'`, stroke color only.
- Changeset bumps `icon-registry`, `react-icon`, `vue-icon` (`minor`), matching the convention from prior icon-add PRs.

Verified: `tsc --noEmit` clean, registry tests 5/5 pass, rendered all three sizes in-browser (consistent replay glyph, `currentColor` cascades), and ran `.agents/review-checklist.md` (no raw hex, iconography attrs correct, changeset present).

## Where to focus review

- **Dropping the disc** on 24px — agree the icon should stay single-tone to match 16px, rather than reintroducing the two-tone disc as a separate variant?
- Is the **12px solid glyph** (vs. stroked) the right call at that size, consistent with how we treat other small icons?

---

> [!NOTE]
> **Low Risk**
> Icon SVG and registry additions only; default transparent disc preserves existing 16px appearance and theming via `currentColor`.
> 
> **Overview**
> Adds **12px** and **24px** variants of `action-test-replay` so consumers can use `IconActionTestReplay` at all three sizes instead of local stand-in SVGs.
> 
> All three sizes (including updated **16px**) gain an optional **`icon-light`** circular background that defaults to **transparent**, so bare `currentColor` glyph usage stays the same; passing **`fillColor`** can show the disc. The **12px** asset is a solid glyph; **16px** and **24px** keep the stroked replay construction. **Minor** changesets cover `@cypress-design/icon-registry`, `react-icon`, and `vue-icon`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cc7baebca626f806cfe604f1c284519b0276077d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


icon-registry/icons-static/action-test-replay_x24.svg
ryanjwilke, line 1: This one is missing the circular fill inside it.

ryanjwilke: Fixed in bbf9090b — added the `icon-light` disc (`<circle r="12" fill="#D0D2E0">`) behind the glyph on the 24px icon, matching the Figma source and the two-tone pattern other `_x24` icons use. Left the 12px as a solid glyph since the Figma export has no disc at that size; metadata now reports `hasFillColor: ["24"]`. Verified the disc renders and tsc/tests pass.

ryanjwilke: Follow-up: per discussion, the disc is now an opt-in fill (transparent by default) at all three sizes (12/16/24), added in cc7baebc. Pass `fillColor` to show it; bare `<IconActionTestReplay />` renders disc-free, so existing usages like TestResult are unchanged.

icon-registry/icons-static/action-test-replay_x12.svg
```diff
@@ -0,0 +1,4 @@
+<svg viewBox="0 0 12 12" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <circle cx="6" cy="6" r="6" fill="transparent" class="icon-light" />
+    <path d="M6 0C9.31371 0 12 2.68629 12 6C12 9.31371 9.31371 12 6 12C4.36562 12 2.88192 11.3455 1.80078 10.2861C1.40632 9.89959 1.39959 9.26556 1.78613 8.87109C2.17265 8.47671 2.80574 8.47099 3.2002 8.85742C3.92267 9.56539 4.90977 10 6 10C8.20914 10 10 8.20914 10 6C10 3.79086 8.20914 2 6 2C4.13616 2 2.57002 3.27477 2.12598 5H2.42871C2.98093 5.00008 3.42871 5.44776 3.42871 6C3.42871 6.55224 2.98093 6.99992 2.42871 7H1C0.447715 7 6.4426e-08 6.55228 0 6C0 2.68629 2.68629 6.44277e-08 6 0ZM4.5 4.40137C4.50002 3.80236 5.16761 3.44508 5.66602 3.77734L8.06348 5.37598C8.50878 5.67284 8.50878 6.32716 8.06348 6.62402L5.66602 8.22266C5.16761 8.55492 4.50002 8.19764 4.5 7.59863V4.40137Z" fill="currentColor" class="icon-dark" />
```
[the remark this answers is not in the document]
ryanjwilke: Verified against a render: the disc does show at 12px when `fillColor` is passed — the gray fills the interior negative space between the ring and the play triangle. What differs from 16/24 is that the 12px ring is a solid annulus (not a 2px stroke), so less of the disc shows and its outer edge sits behind the ring. That's the single-path Figma-glyph tradeoff, documented in cc7baebc and intentional; the disc is opt-in (transparent by default), so bare usages are unaffected. Closing as out-of-scope — no change.

PR #689, opened 2026-06-17
feat(runresults): styling fixes + native 12px self-healed icon

emilmilanov:
## What & why

Styling and correctness fixes for `RunResults`, surfaced while integrating it into cypress-services (replacing the in-house `RunStats`), plus a native 12px self-healed icon.

### RunResults
- **Border** is now drawn as an absolutely-positioned `::after` overlay instead of a plain border / inset shadow on the `<ul>`. This:
  - fixes the pill height regressing to 26px → back to **24px** (border no longer adds to box height),
  - keeps the border **visible when a stat is hovered** (the overlay paints above the link backgrounds, which previously covered an inset border/shadow),
  - **renders even with Tailwind `preflight` disabled** (consumers like cypress-services disable preflight).
- **Links** keep gray text and **no underline on hover** (overrides host `a:hover` styles such as Bootstrap's `$link-color`).
- `renderLink` now receives the computed link **`className`** so custom (e.g. router) links match the default `<a>` styling.
- New **`pillClassName`** prop (merged onto the `<ul>` via **`tailwind-merge`**) to override the pill background — e.g. `bg-gray-900` on a dark surface, or `bg-transparent`. `className` stays on the **root wrapper** `<div>` (kept intentionally — it will hold multiple stat lists in the future).

### Icons
- Add **`general-sparkle-single`** as a native **12px** icon and use it for the self-healed stat, replacing a scaled-down 16px icon.

### Build / packaging
- Collapse `constants-runresults` into the React/Vue packages: it's now **private** and bundled in (JS inlined from devDeps; self-contained `.d.ts` via `rollup-plugin-dts` / `vite-plugin-dts`), so consumers install a **single package** per framework. `tailwind-merge` is a new (externalized) dependency. The vue UMD-globals map is sourced from each package's `build.lib.name`.

## Changesets
- `@cypress-design/{react,vue}-runresults`: minor (`constants-runresults` is now private — not published)
- `@cypress-design/{icon-registry,react-icon,vue-icon}`: minor

## Verification
- React component tests / screenshots confirm the 24px height, visible-on-hover border, gray no-underline hover links, and the 12px self-healed icon.
- `pillClassName` verified end-to-end in the cypress-services dashboard (dark RunResults pill set to `bg-gray-900`, single border).

---

> [!NOTE]
> **Medium Risk**
> Minor API change to `renderLink` and new `pillClassName` may require consumer updates; packaging shifts how types/constants ship but is intentional for single-package installs.
> 
> **Overview**
> **RunResults** (React + Vue) gets integration-focused styling and API tweaks: the pill border is drawn with an inset **`::after` shadow** instead of a real border (restores **24px** height, stays visible on stat hover, works without Tailwind preflight). Link theme classes add **gray text and `hover:no-underline`**. **`renderLink`** now receives a third **`className`** argument so router links match the default anchor. **`className`** targets the root wrapper; new **`pillClassName`** merges onto the `<ul>` via **`tailwind-merge`** so backgrounds like `bg-gray-900` override the theme.
> 
> The self-healed stat switches from a scaled 16px icon to native **`IconGeneralSparkleSingle`** at 12px; **`general-sparkle-single_x12.svg`** is added to the icon registry.
> 
> **Packaging:** `@cypress-design/constants-runresults` is **private** and inlined into published JS/types so consumers install only `@cypress-design/react-runresults` or `vue-runresults` (React uses `rollup-plugin-dts`; Vue uses `vite-plugin-dts` with `bundledPackages`). Shared **`vue.vite.config`** gains configurable externals/plugins and UMD global mapping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed73547037f6ca46cacb4923eff41b1fa3c95966. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


components/RunResults/constants/src/index.ts
```diff
@@ -25,8 +25,15 @@ export type LeadingStatKey = (typeof LeadingStatKeys)[number]
 export const CssClasses = {
   // Outer wrapper. `inline-flex` so the component shrinks to content width.
   container: 'inline-flex pointer-events-auto',
-  // The <ul> pill itself. Theme overrides border and text colors via CssTheme.
-  list: 'flex items-center text-[14px] leading-[24px] font-medium list-none border rounded-[4px]',
+  // The <ul> pill itself. Theme provides the border color and text colors via
```
ryanjwilke: I find that claude code is adding these very specific inline comments far too much. We don't always need to know every single reason for why one utility class is used over another. I would recommend removing this comment.

It's better to add context (when needed) to some kind of agent file that is specifically for this component and for AI agents to read through when updating or using the component.

components/RunResults/constants/src/index.ts
```diff
@@ -40,27 +47,24 @@ export const CssClasses = {
...
   // IconStatusFlaky is unchanged.
   iconFlaky: 'mx-[4px] [&_path:first-child]:fill-transparent',
-  // Self-healed icon override — `IconGeneralSparkleSingleSmall` only ships
-  // a `["16"]` variant in the icon registry, but the rest of the stats
-  // render at 12px. The icon component IS an <svg>, so `w-3 h-3` on the
-  // className overrides the icon's intrinsic `width`/`height` attributes
-  // via CSS and pins the rendered size to 12 × 12 — visual consistency
-  // without depending on a 12px icon variant that doesn't exist.
-  iconSelfHealed: 'mx-[4px] w-3 h-3',
+  // Self-healed icon margin — matches the other stat icons. Uses the native
```
ryanjwilke: This is another comment that isn't needed. All these comments are burning through tokens every time an agent opens this file up. It doesn't need this kind of context each time.

components/RunResults/react/RunResults.tsx
```diff
@@ -199,7 +201,16 @@ export const RunResults = React.forwardRef<
       data-cy="run-results"
       className={clsx(CssClasses.container, className)}
     >
-      <ul className={clsx(CssClasses.list, CssTheme[theme].list)}>
+      <ul
+        className={clsx(
+          CssClasses.list,
+          // bgClassName replaces the theme's default background so a consumer
+          // can blend the pill with a colored surface.
+          bgClassName
```
ryanjwilke: Couldn't this just be part of the component where you pass a `className="bg-red-500"` where needed? This wouldn't require an extra `bgClassName` property as part of the API of this component.

components/RunResults/vue/vite.config.ts
```diff
@@ -1,7 +1,18 @@
...
-export default generateViteConfig({
-  entry: path.resolve(__dirname, './index.ts'),
-  name: 'RunResults',
-})
+export default generateViteConfig(
+  {
+    entry: path.resolve(__dirname, './index.ts'),
+    name: 'RunResults',
+  },
+  // Externalize the workspace deps this component imports so they aren't
```
ryanjwilke: Again, do these comments need to be inlined here? Is this useful context every time a component opens this file or is this just useful this one time? Claude does this too much, so I have to ask.

ryanjwilke:
I see Vue/React tabs, but that doesn't appear to change anything about the **Install** section for example which just has standard notes that kind of blend all component frameworks together (see attached screenshot).
Screenshot 2026-06-17 at 22 22 38

ryanjwilke:
## Review — thanks @emilmilanov 🙏

> **Update (Jun 18):** Emil addressed the whole checklist in `7e48077b` — verified against CI (`Build`, `Template Test`, and the React/Vue build jobs all green, so the bundled types resolve). Everything below is checked off **except the `bgClassName` API shape**, which is still open — consolidated recommendation in the thread [here](https://github.com/cypress-io/cypress-design/pull/689#discussion_r3437128396).

### 🔴 Blocker — API design
- [ ] **`bgClassName` vs `className` — still open.** Recommendation posted on the thread: merge the wrapper so the `<ul>` is the root, drop `bgClassName`, and route `className` through `tailwind-merge` so a consumer's `bg-*` reliably wins (enum ruled out — the real values, `bg-gray-900` / `bg-transparent`, are arbitrary host-surface colors, not a semantic set). `pillClassName` + the existing `listClasses` replace is the dep-free fallback. Awaiting decision.

### 🟠 Recommended in this PR — collapse `constants-runresults` into the component packages ✅
- [x] Constants bundled into both builds; package marked `"private": true`, changeset removed.
- [x] `.d.ts` made self-contained via `rollup-plugin-dts` (React) + `vite-plugin-dts`/api-extractor (Vue) — `bundledPackages`/inlined types so consumers never install `constants-runresults`. **Verified:** `Build` + `Template Test` pass, so the bundled types resolve for a real consumer.

### 🟠 Code quality ✅
- [x] **De-duplicated the `bgClassName` swap** into a shared `listClasses(theme, bgClassName?)` helper in constants, used by both frameworks (with the no-`bg-*` fallback).
- [x] **UMD global derivation fixed** — replaced the `Cy*` auto-derive with an explicit `UMD_GLOBALS` map keyed on each package's `build.lib.name`, and it now throws if an externalized dep has no mapped global (can't silently drift again).
- [x] **Inline comments trimmed; rationale moved to `architecture.md`.**

### 🟡 Tests ✅
- [x] `renderLink` className passthrough asserted (React `RunResultsReact.cy.tsx:47`, Vue `:46`) — checks the 3rd arg arrives and yields `no-underline` / `px-[6px]`.
- [x] `bgClassName` override test (`assertions.ts:411`, shared across React + Vue) — swaps the `bg-*` token, keeps other theme classes intact.
- [x] 12px self-healed icon asserted (`assertions.ts:442`) — "renders the native 12px icon (no `w-3 h-3` size override)."

### 🟡 Docs ✅
- [x] `renderLink` updated to `(href, children, className?)` in `instructions.md` + `architecture.md`.
- [x] `bgClassName` documented in `instructions.md`. *(Note: if we go with the merge + `tailwind-merge` recommendation, these `bgClassName` docs will need a follow-up edit.)*

### ✅ Resolved (fixed in `37bb491b`)
- [x] Bugbot — "Vue `renderLink` omits className".
- [x] Bugbot — "Vue ignores `bgClassName`".

### ⚪ Optional / minor
- [ ] Icon taxonomy: `general-sparkle-single` exists only at 12px while its `-small`/`-large` siblings exist at 16/24. Fine for this use — flagging for icon-registry consistency.

components/RunResults/react/RunResults.tsx
emilmilanov: Kept `bgClassName` as a dedicated prop — here's the reasoning after digging into the conventions:

- In this DS, `className` already means **append to the component's root element** (see Button/Textbox). RunResults' root is the outer container, not the pill `<ul>`. Repointing `className` to the pill would break that convention **and** break a current consumer (cypress-services) that styles the container via `className`.
- Routing arbitrary classes to the pill also needs conflict resolution: a second `bg-*` loses to the theme's by Tailwind source order, so a general escape hatch would require `tailwind-merge` — which isn't in the repo, and adding it DS-wide for one property felt like the wrong call here.

So `bgClassName` keeps `className` conventional and stays dependency-free. The swap is now de-duped into a shared `listClasses(theme, bgClassName?)` helper (constants), used by both frameworks. Happy to revisit a general pill escape hatch DS-wide if a second inner-override need shows up. (7e48077b)

components/RunResults/constants/src/index.ts
emilmilanov: Trimmed — moved the rationale to `architecture.md` (the designated home per AGENTS.md). (7e48077b)

components/RunResults/constants/src/index.ts
emilmilanov: Trimmed — moved the rationale to `architecture.md` (the designated home per AGENTS.md). (7e48077b)

components/RunResults/vue/vite.config.ts
emilmilanov: Trimmed — moved the rationale to `architecture.md` (the designated home per AGENTS.md). (7e48077b)

components/vue.vite.config.ts
```diff
@@ -3,8 +3,34 @@ import { defineConfig } from 'vite'
...
 
-export default (libConfig: LibraryOptions) =>
-  defineConfig({
+// Derive a UMD global name for an externalized @cypress-design package, e.g.
+// '@cypress-design/constants-runresults' -> 'CyConstantsRunresults'.
+const umdGlobal = (name: string) =>
+  'Cy' +
+  name
+    .replace('@cypress-design/', '')
+    .replace(/(^|-)(\w)/g, (_m, _d, c) => c.toUpperCase())
```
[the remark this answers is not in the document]
emilmilanov: Fixed — replaced the `Cy*` auto-derivation with an explicit, verified map keyed by each package's `build.lib.name` (`vue-tooltip` → `Tooltip`, `vue-statusicon` → `StatusIcon`, `vue-icon` → `Icon`, `icon-registry` → `CypressIconRegistry`). It now also **throws** if an externalized dep has no mapped global, so this can't silently drift again. (7e48077b)

components/RunResults/constants/src/index.ts
```diff
@@ -147,11 +151,16 @@ export interface RunResultsProps {
...
+  // renders for the inner icon + count. `className` is the component's computed
+  // link styling (gray text, padding, no-underline, hover/focus) — apply it to
+  // your custom link so it matches the default <a>.
+  renderLink?: (href: string, children: unknown, className?: string) => unknown
 
   showTooltip?: boolean
   className?: string
+  // Override the pill's background (e.g. `bg-transparent` to blend with a
+  // colored surface). Replaces the theme's default background on the <ul>.
+  bgClassName?: string
```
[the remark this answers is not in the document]
emilmilanov: Done — `instructions.md` and `architecture.md` now document the `(href, children, className?)` signature, and `bgClassName` is documented as a new prop. (7e48077b)

components/RunResults/react/RunResults.tsx
```diff
@@ -93,7 +94,7 @@ const Stat: React.FC<StatProps> = ({
   let content: React.ReactNode
   if (isLinked && link) {
     if (renderLink) {
-      content = renderLink(link, inner) as React.ReactNode
+      content = renderLink(link, inner, linkClasses) as React.ReactNode
     } else {
```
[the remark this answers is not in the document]
emilmilanov: Added — `passes the computed link className to renderLink` asserts the 3rd arg is received and, when applied, yields the default link styling (`no-underline`, `px-[6px]`). (7e48077b)

components/RunResults/react/RunResults.tsx
```diff
@@ -199,7 +201,16 @@ export const RunResults = React.forwardRef<
...
-      <ul className={clsx(CssClasses.list, CssTheme[theme].list)}>
+      <ul
+        className={clsx(
+          CssClasses.list,
+          // bgClassName replaces the theme's default background so a consumer
+          // can blend the pill with a colored surface.
+          bgClassName
+            ? CssTheme[theme].list.replace(/\bbg-\S+/, bgClassName)
+            : CssTheme[theme].list,
+        )}
```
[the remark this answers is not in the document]
emilmilanov: Added — a `background override` test asserts `bgClassName` swaps the `bg-*` token and leaves the other theme classes (`text-*`, the `after:` border shadow) intact. (7e48077b)

components/RunResults/vue/RunResults.vue
```diff
@@ -150,16 +152,20 @@ export default defineComponent({
 
       let content: VNode
       if (link) {
+        const linkClasses = joinClasses(
+          CssClasses.link,
+          CssTheme[props.theme].link,
+        )
         if (props.renderLink) {
-          content = props.renderLink(link, inner) as VNode
+          content = props.renderLink(link, inner, linkClasses) as VNode
```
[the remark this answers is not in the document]
emilmilanov: Added the same className-passthrough assertion on the Vue side. (7e48077b)

components/RunResults/vue/RunResults.vue
```diff
@@ -306,7 +312,17 @@ export default defineComponent({
...
-              class: joinClasses(CssClasses.list, CssTheme[props.theme].list),
+              class: joinClasses(
+                CssClasses.list,
+                // bgClassName replaces the theme's default background so a
+                // consumer can blend the pill with a colored surface.
+                props.bgClassName
+                  ? CssTheme[props.theme].list.replace(
+                      /\bbg-\S+/,
+                      props.bgClassName,
+                    )
```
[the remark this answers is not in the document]
emilmilanov: Covered — the `bgClassName` override test lives in the shared `assertions.ts`, so it runs for both React and Vue. (7e48077b)

components/RunResults/vue/RunResults.vue
```diff
@@ -122,7 +122,8 @@ export default defineComponent({
         })
       }
       if (statKey === 'selfHealed') {
-        return h(IconGeneralSparkleSingleSmall, {
+        return h(IconGeneralSparkleSingle, {
+          size: '12',
```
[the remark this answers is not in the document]
emilmilanov: Fixed in 37bb491b.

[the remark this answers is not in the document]
emilmilanov: Fixed in 37bb491b.

emilmilanov:
Thanks for the thorough review @ryanjwilke (and @Copilot) 🙏 — worked through the whole checklist. Pushed as `7e48077b` (the two Bugbot items were already in `37bb491b`).

### 🔴 Blocker — `bgClassName` vs `className`
Kept `bgClassName`. In this DS `className` already means *append to the root element* (Button/Textbox), and RunResults' root is the container, not the pill — repointing it would break that convention and a current cypress-services consumer. A general pill escape hatch would also need `tailwind-merge` (not in the repo) to win the `bg-*` source-order conflict. So: narrow, dependency-free prop, with the swap de-duped into a shared `listClasses()` helper. Details in the inline thread — glad to revisit DS-wide if a second inner-override need appears.

### 🟠 Collapse `constants-runresults` — done in this PR
- Now `"private": true`, moved to `devDependencies` of both packages, dropped from react `external` / vue `extraExternal`, removed from the changeset → JS is bundled in.
- Self-contained types: `rollup-plugin-dts` (react) + `vite-plugin-dts` `rollupTypes` (vue). Verified **both** `dist/index.d.ts` have **zero** `constants-runresults` references; the real runtime deps stay external.
- Consumers now install a single package per framework.

### 🟠 Code quality
- **De-dup** — `replace(/\bbg-\S+/, …)` is gone; both frameworks call `listClasses(theme, bgClassName?)` (tolerant of a theme with no `bg-*`).
- **UMD globals** — explicit verified map keyed by `build.lib.name`, throws on a missing entry.
- **Comments** — trimmed; rationale moved into `architecture.md`.

### 🟡 Tests
Added (React + Vue): `renderLink` className passthrough, `bgClassName` override (swaps bg, keeps other theme classes), and the 12px self-healed icon. Also **fixed a stale dark-theme assertion** that still checked `border-gray-800` (now an `::after` shadow). All 56 RunResults CT + 10 a11y specs pass.

### 🟡 Docs
`instructions.md` + `architecture.md`: `(href, children, className?)` signature, `bgClassName` prop, single-package Install (matches Tag/Tooltip), updated type-export wording.

### ⚪ Icon taxonomy
Noted — `general-sparkle-single` exists only at 12px for now; flagging for a separate icon-registry consistency pass.

---
Also dogfooded the result in the real consumer (cypress-services dashboard): the collapsed single package renders, the dark pill honors `bgClassName="bg-gray-900"`, the self-healed icon is a native 12px, and the bundled Tailwind classes generate correctly.

components/RunResults/react/RunResults.tsx
ryanjwilke: Okay, here's my thoughts on this:

1. I would not expect `className` and `bgClassName` to be targeting two separate components. Typically you would see something like this on other components: `className` (on the root container) `listClassName` on the container that you're aiming to style.
2. Another option would be to just pass something a bit more fixed like `background='light'` or something like that. This sends the impression that you maybe have a light/dark theme. It could also be combined with a themed style. At least that kind of accomplishes it without needing to offer a single property for just background classes which seems a bit strange compared to other solutions I've seen in the past.

components/RunResults/react/RunResults.tsx
ryanjwilke, line 197: Is there a need for a separate "container" `div` if the `ul` could act as a container? This could also fest the `className` vs `bgClassName` properties confusion that exists as that API isn't very intuitive today.

If the `ul` could be the container and the design still holds together it makes for less code which isn't bad, and could simplify the component API.

components/RunResults/react/RunResults.tsx
ryanjwilke: Here's my recommendation to close this out.

Make the `<ul>` the root element and remove the wrapper `<div>`. The wrapper used to draw the border, but this PR already moved the border onto the `<ul>`, so the wrapper no longer does anything we need. With the `<ul>` as the root, a single `className` lands right on the pill — so we don't need a separate `bgClassName`, and the confusing split between two class props goes away.

One thing to handle: a `bg-*` class you pass can lose to the theme's own background, since both set a background. Adding `tailwind-merge` (a small, common library) fixes this — it makes the class you pass win:

```
twMerge(CssClasses.list, CssTheme[theme].list, className)
```

Then `className="bg-gray-900"` or `className="bg-transparent"` works as expected, and the border is left alone.

Result: one `className` on the pill, no `bgClassName`, one less element. RunResults has no consumers yet, so this is a safe time to make the change.

emilmilanov: Done — went with your primary recommendation in `057620e3`.

- **Merged the wrapper away.** The pill `<ul>` is now the root: `ref` (`HTMLUListElement`), `data-cy="run-results"`, passthrough attrs, and `className` all land on it. The container `<div>` and `CssClasses.container` are gone.
- **Dropped `bgClassName`** (and the `listClasses` helper).
- **`className` → `twMerge(CssClasses.list, CssTheme[theme].list, className)`**, so `bg-gray-900` / `bg-transparent` win the source-order conflict; the `::after` border and text size are untouched. Vue normalizes a string/array/object fallthrough `class` with `clsx` before `twMerge`. `tailwind-merge` is a real dep of both packages (externalized like `clsx`).
- **Tests/docs:** selectors target the root `<ul>`; the old `bgClassName` test is now a `className`-override test; a11y still green (selectors are `[data-cy="run-results"] a, button`). architecture.md + instructions.md updated; `RunResults has no consumers` so the `div`→`ul` ref change was free.

All 56 RunResults CT + 10 a11y specs pass. Also dogfooded in cypress-services: the dark run-tooltip pill now uses `className="bg-gray-900"`, renders a single border, and the consumer's outer spacing moved to a wrapper. Thanks for pushing on this — the merged API is cleaner.

emilmilanov: Update — I need to walk back the wrapper merge from `057620e3`. New context that wasn't on the table: **the wrapper `<div>` is not vestigial — it will hold multiple `<ul>` stat lists in the near future** (today it's just the one). So removing it would've forced us to re-add it shortly.

Landed in `38cd3ceb`, going with the conventional two-prop split (which is actually your *first* suggestion on this thread):

- **`className` → root wrapper** (`clsx`, DS convention — Button/Textbox); `ref` back to `HTMLDivElement`, `data-cy` + passthrough attrs on the wrapper.
- **`pillClassName` → the `<ul>`**, merged via `twMerge(CssClasses.list, CssTheme[theme].list, pillClassName)` — so `bg-gray-900` / `bg-transparent` still win the source-order conflict. `tailwind-merge` stays (it's what powers this).

So the "two class props target two elements" concern is addressed by naming them for their targets, and we keep the wrapper for the upcoming multi-list layout. Tests (selectors back to `[data-cy="run-results"] ul`, `pillClassName` override test), docs, and the changeset are updated; 56 CT + 10 a11y green. Dogfooded in cypress-services: dark run-tooltip uses `pillClassName="bg-gray-900"` on the pill + `className` for the wrapper's outer spacing, single border, renders correctly.

.changeset/runresults-styling-fixes.md
```diff
@@ -0,0 +1,6 @@
+---
+'@cypress-design/react-runresults': minor
+'@cypress-design/vue-runresults': minor
+---
+
+RunResults: draw the border as an inset `::after` overlay (fixes the 26px height back to 24px, keeps the border visible on stat hover, and renders without Tailwind preflight), keep link text gray with no underline on hover, and expose the link `className` to `renderLink` so router links match the default anchor. `className` applies to the root wrapper (DS convention); a new `pillClassName` applies to the `<ul>` and is merged via `tailwind-merge`, so a consumer can override the pill background (e.g. `bg-gray-900` / `bg-transparent`) and win the Tailwind source-order conflict. The self-healed stat now uses the native 12px `general-sparkle-single` icon. The `constants-runresults` package is bundled into the React and Vue packages — consumers install a single package.
```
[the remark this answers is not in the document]
emilmilanov: Good catch — updated the PR description to use `pillClassName` (and refreshed the Build/Changesets sections). The code, docs, and changeset were already consistent on `pillClassName`; only the description lagged.

ryanjwilke:
Latest review — `38cd3ceb`. The `pillClassName` split is the right call, and the build/dep handling checks out. Everything is green; nothing here blocks merge. Four small items, ranked — fix the first two before merge, the last two are optional.

**1. Pin `tailwind-merge` to a real range (not `"*"`)** — `components/RunResults/react/package.json` and `components/RunResults/vue/package.json`.
`tailwind-merge`'s class knowledge is tied to the Tailwind major version, so `"*"` can resolve a version that doesn't match a consumer's Tailwind and silently merge wrong. Pin it to the major that matches this repo's Tailwind (check the root `tailwindcss` version and set e.g. `^x.0.0`). `clsx` can stay `"*"` — it's version-agnostic.

**2. Fix the stale PR description** — the "What & why" and "Verification" sections still say `bgClassName`; the shipped prop is `pillClassName`. Update both mentions so the release notes match the API.

**3. (Optional) Add a test that `className` lands on the wrapper, not the `<ul>`.** `assertions.ts` tests `pillClassName` on the `<ul>`, but nothing asserts `className` applies to the root wrapper. A small assertion locks in the two-prop contract.

**4. (Optional) Drop the redundant `clsx` wrap in Vue.** `vue/RunResults.vue` calls `twMerge(clsx(...))`; React calls `twMerge(...)` directly. `twMerge` already takes multiple args and ignores falsy, so the `clsx` wrap isn't needed — minor parity cleanup.

emilmilanov:
Thanks — all four addressed in `a7a55213`.

1. **`tailwind-merge` pinned to `^2.0.0`.** Good catch, and it was already live: `"*"` had resolved **3.6.0** (built for Tailwind v4's class groups) while this repo is on **Tailwind 3.4.3** — exactly the silent-mismatch you flagged. `^2` now resolves **2.6.1**. `clsx` left at `"*"` (version-agnostic).
2. **PR description** — fixed (the "What & why"/"Verification" `bgClassName` mentions → `pillClassName`, plus refreshed the Build/Changesets sections). Landed just before your review came in.
3. **Added** a `class targeting` test: `className` lands on the wrapper (and not the `<ul>`), `pillClassName` on the `<ul>` (and not the wrapper) — locks in the two-prop contract.
4. **Dropped** the redundant `clsx` wrap in Vue (`twMerge(...)` directly now, matching React) and removed the now-unused `clsx` dep + externalization.

All green — 58 RunResults CT + 10 a11y.

ryanjwilke (approved):
This version looks good to me. Nice work 👍🏽

PR #723, opened 2026-09-18
feat(favicon): add the favicon package, its components, and a docs page

emilmilanov:
## What & why

Every Cypress web surface ships the same broken favicon, and each repo carries its own copy of it.
`cypress.io`, `cypress-design/docs` and `cypress-services` ship a **byte-identical** `favicon.ico`
(md5 `15c97efc…`) that is not an ICO at all — a bare 48×48 PNG wearing an `.ico` extension. Every
frame across every repo has a **fully transparent centre**, so the navy `#1B1E2E` wordmark sits on
nothing and is close to invisible on Chrome's dark tab strip.

Worse, and only findable by testing: **Chrome was never using our adaptive SVG.** The widely copied
`<link rel="icon" href="/favicon.ico" sizes="any">` recipe — which cypress.io and
cypress-documentation both ship — makes Chrome prefer the ICO and never even download the SVG.

This adds a package so the artwork and, more importantly, the *markup rule* live in one place
instead of being re-derived (and re-broken) in four repos. It also fixes this site's own favicon,
which was the worst case of the bug it now documents.

## What's in it

| | |
| --- | --- |
| `@cypress-design/favicon` | The six asset files, `FAVICON_LINKS`, `MANIFEST_ICONS`, `faviconHeadTags()`, `faviconLinksHtml()`, and a `cypress-favicon <dir>` bin that copies the assets into a site's static directory |
| `@cypress-design/vue-favicon` | `<Favicon />` — renders the head tags. For cypress.io |
| `@cypress-design/react-favicon` | The same component for React-registered sites. For this site |
| `/favicons` docs page | The asset set rendered at real size, plus the reasoning |

A favicon is fetched by the browser as a plain URL before any JavaScript runs, so unlike every other
package here this one emits real files rather than only a module.

## The measurement behind the markup

Five markup variants served, recording which file each engine actually fetched. **Chrome
153.0.8010.48** and **Safari 26.5**:

| Markup | Chrome 153 | Safari 26.5 | Fetched |
| --- | --- | --- | --- |
| SVG first, `sizes="any"` on ICO | ICO | ICO | **ICO only** |
| `sizes="any"` on both | ICO | ICO | **ICO only** |
| Explicit sizes on the ICO | ICO | ICO | **ICO only** |
| ICO first, SVG last | ICO | ICO | both |
| **SVG only, no ICO link** | **SVG, adapts** | **SVG** | SVG |

So `FAVICON_LINKS` omits two tags that are widely recommended:

- **No `rel="icon"` for the ICO.** A declared ICO beats the SVG in both engines regardless of
  attributes or order. The ICO is still shipped and still served — clients that want it request
  `/favicon.ico` from the origin root by convention, without reading the markup.
- **No `rel="mask-icon"`.** Safari stopped requiring a monochrome pinned-tab SVG in Safari 12 (2018);
  verified inert in Safari 26.5, where pinning a tab renders the full-colour favicon either way.

## Why two component packages rather than one

Both builds render identically, but using the wrong one does not merely duplicate — it breaks the
build. Astro resolves a component's renderer by asking each registered one in turn, and
`@astrojs/react`'s check reads `Component["$$typeof"].toString()` for any object. A compiled Vue SFC
is a plain object with no `$$typeof`, so it throws rather than returning false. Verified both ways:

- **cypress.io** registers Vue only → `vue-favicon` renders clean.
- **This site** registers React then Vue → `react-favicon` renders clean; `vue-favicon` 500s every
  MDX page.

Docs and Cloud use the functions instead: a Docusaurus plugin returns tag descriptors before any
component tree exists, and Cloud's auth views ship zero `<script>` tags so nothing can mount there.

## Assets

Generated from one adaptive source SVG. `favicon.ico` is a real container with five PNG frames
(16/24/32/48/64) baked from the **dark** variant with the disc filled — it cannot adapt, so it has to
be one image that survives on either tab strip. Frames carry an unsharp mask (edge energy at 16px
160 → 230). Larger rasters are supersampled 4×, because the ring is four separate quarter paths whose
independently antialiased edges left a visible seam at the cardinal points — 25.5/255 deep at 512px.

## Where to focus review

- **Should the design system emit files at all?** This is the first package that does; every other
  one compiles to a JS payload. It sets a precedent for logos, OG images and similar.
- **The two omitted tags.** They are the whole point of the package. The measurements are
  reproducible if you want to check them.
- `build:components` needed an explicit filter for `packages/favicon`: its turbo filters cover
  `constants-*`, `vue-*` and `react-*`, and that script *is* the root `postinstall`, so a `packages/`
  workspace with no dependents would silently publish an empty `dist`.

## Verified

Dev servers for all three sites: head markup correct, every asset 200, `/favicon.ico` serving the
5-frame container, and `safari-pinned-tab.svg` correctly 404ing on cypress.io. The `/favicons` page
has a component spec covering the asset list, image decoding, and a Percy snapshot.

## Known limitation

Safari renders the SVG but never evaluates `prefers-color-scheme` — it always resolves light. A
Safari dark-mode tab shows the light variant while Chrome and Firefox show the dark one. Legible,
since the white disc still gives the mark a background; accepted and documented on the page.

---

> [!NOTE]
> **Low Risk**
> Changes are mostly new packages, static assets, and docs head markup—not auth, payments, or core app logic. Wrong Astro framework choice for `<Favicon />` can break builds, which is documented.
> 
> **Overview**
> Introduces a **central favicon stack** so Cypress sites share one asset set and one head-markup policy instead of duplicating broken ICO/SVG recipes.
> 
> **`@cypress-design/favicon`** ships the icon files, `FAVICON_LINKS` / `MANIFEST_ICONS`, `faviconHeadTags()` and `faviconLinksHtml()` for non-component heads, and a **`cypress-favicon`** CLI to copy assets into static dirs. **`FAVICON_LINKS` intentionally skips a declared ICO link and `mask-icon`**, so adaptive SVG is used in modern browsers while `/favicon.ico` stays at the root for legacy clients.
> 
> **`@cypress-design/react-favicon`** and **`@cypress-design/vue-favicon`** server-render the same `<link>` tags (optional `links` prop). Docs adopt **`react-favicon`** in `BaseLayout.astro` (manifest tag filtered out) and **`build:docs`** copies package assets into `docs/public`.
> 
> Adds a **`/favicons`** page with a **`FaviconAssets`** showcase, Cypress/Percy coverage, component docs, CI cache paths for `packages/favicon/dist`, and **`build:components`** / root dependency wiring so the favicon package builds on install.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8d82d44e80b5c6c5982d8f620ef2e1c5d26b15a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/src/FaviconAssets.vue
```diff
@@ -0,0 +1,210 @@
...
+    alpha: 'yes',
+    used: 'Android, installed PWAs',
+  },
+  {
+    name: 'android-chrome-512x512.png',
+    dims: '512×512',
+    alpha: 'yes',
+    used: 'Android splash screens',
+  },
+]
```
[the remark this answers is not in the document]
emilmilanov: Correct, and fixed in 7f2fd05. The rows are now generated from `FAVICON_ASSETS` with the prose held in a lookup keyed by filename, so an asset added or removed is reflected automatically.

The same drift existed in `bin/copy-assets.mjs`, which read the assets directory rather than the declared list — a stray file would have been copied into a consumer's static directory. That now copies `FAVICON_ASSETS` too.

docs/src/FaviconAssets.cy.ts
```diff
@@ -0,0 +1,42 @@
+/// <reference types="cypress" />
+import { mount } from 'cypress/vue'
+import { FAVICON_ASSETS } from '@cypress-design/favicon'
+import FaviconAssets from './FaviconAssets.vue'
+
+describe('FaviconAssets', () => {
+  beforeEach(() => {
+    mount(FaviconAssets)
```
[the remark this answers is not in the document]
emilmilanov: Good catch — fixed in 22c581f. `packages/favicon/dist` is now in every cache path list (8 of them, across build/test/e2e-tests/template-test), sitting beside `packages/details-animation/dist` which is there for exactly this reason.

You were right about the mechanism: the job restores node_modules plus the built dists and skips `yarn install` on a cache hit, so nothing rebuilds the workspace — and since the package's `exports` point only at `dist/`, the spec would have failed to resolve the module on any cached run.

components/Favicon/vue/vite.config.ts
```diff
@@ -0,0 +1,7 @@
+import * as path from 'path'
+import generateViteConfig from '../../vue.vite.config'
+
+export default generateViteConfig({
+  entry: path.resolve(__dirname, './index.ts'),
+  name: 'Favicon',
+})
```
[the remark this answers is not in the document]
emilmilanov: Right, and a real one — fixed in the commit above.

The vue build now externalizes `@cypress-design/favicon` via `extraExternal`, with a UMD global registered in `vue.vite.config.ts`. Verified: the vue bundle no longer contains the link hrefs and imports `FAVICON_LINKS` at runtime, matching what the react build was already doing through `Object.keys(pkg.dependencies)`.

Your framing was the useful part — a favicon-only release would have left a stale array baked into vue-favicon while react-favicon picked up the new one, breaking the one thing the package is supposed to guarantee.

ryanjwilke:
@emilmilanov A few things jump out at me right away:

1. We could probably make it a bit clearer at the top of the page that there are 3 separate pages for this component (`@cypress-design/favicon`, `@cypress-design/vue-favicon`, `@cypress-design/react-favicon`). Which do I install? Does installing one give me access to both Vue and React components?
2. I feel like we should aim to be deliberate about avoiding language that is specific to each Cypress product or repository. I think it'd be better to split the usage based on the type of framework (i.e. Astro usage, Docusaurus usage, etc). It's probably best that we the documentation clean since this is a public resource that anyone can see.
3. We might want to put this page under the Components section since it does care a Favicon component with it.

.github/workflows/build.yml
```diff
@@ -55,4 +55,5 @@ jobs:
             ./icon-registry/dist
             ./components/**/dist
```
ryanjwilke: If we put this under components, I feel like it could already be covered under: `./components/**/dist` as it would be likely located at `./components/favicon/dist`

.github/workflows/e2e-tests.yml
```diff
@@ -32,6 +32,7 @@ jobs:
             ./icon-registry/dist
             ./components
             ./packages/details-animation/dist
+            ./packages/favicon/dist
```
ryanjwilke: Same comment here...

.github/workflows/template-test.yml
```diff
@@ -35,6 +35,7 @@ jobs:
             ./icon-registry/dist
             ./components
             ./packages/details-animation/dist
+            ./packages/favicon/dist
```
ryanjwilke: Same comment here.

docs/public/apple-touch-icon.png
ryanjwilke: Is it intended to have transparent gaps on the left and right of this image? Perhaps it's an issue with GitHub's display, but it's showing a 1-2px gap on the left and right side in the preview. Just checking.

docs/src/layouts/BaseLayout.astro
```diff
@@ -30,14 +30,17 @@ const frontmatterTocHeadings = (Astro.props as { frontmatter?: { tocHeadings?: t
 const outlineHeadings = tocHeadings ?? frontmatterTocHeadings ?? headings
 const pageTitle = title ? `${title} — Cypress Design System` : 'Cypress Design System'
 const currentPath = Astro.url.pathname
+import Favicon from '@cypress-design/react-favicon'
+import { FAVICON_LINKS } from '@cypress-design/favicon'
```
ryanjwilke: Is this line required in all Astro projects or just this one? I don't remember the documentation showing that.

docs/src/layouts/BaseLayout.astro
```diff
@@ -30,14 +30,17 @@ const frontmatterTocHeadings = (Astro.props as { frontmatter?: { tocHeadings?: t
...
 ---
 
 <html lang="en" data-fw="vue">
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width" />
     <title>{pageTitle}</title>
-    <link rel="shortcut icon" href="/favicon.ico" />
+    {/* This site ships no web manifest, so drop that one tag. */}
```
ryanjwilke: These types of comments don't really feel necessary. If it's important, they should be stored in some kind of architecture.md file or readme next to the `astro` template along with any other requirements or known issues for this astro template.

packages/favicon/assets/apple-touch-icon.png
ryanjwilke: I'm seeing the same issue on this image too.

ryanjwilke: Also, why do we need duplicate versions of the same icon? It feels a bit odd that we'd need that.

AGENTS.md
```diff
@@ -44,6 +44,7 @@ Page-specific acceptance criteria and known issues live in `/.agents/pages/<page
 Available pages:
 
 - `pages/colors.md` — `/colors` page sections, palette behavior, TOC, known issues
+- `pages/favicons.md` — `/favicons` page, the favicon package, and why two common link tags are omitted
```
ryanjwilke: Should we move this to:
`pages/components/favicons.md`

.agents/pages/favicons.md
ryanjwilke: It'd be nice if this could live side-by-side with the template itself if this is for the documentation page(s)

ryanjwilke (changes requested):
I left a few comments for you to consider. Feel free to have Claude do some inline replies if there's some kind of context that I'm missing.

.github/workflows/build.yml
emilmilanov: Good catch, half of it. The component builds do land at `components/Favicon/{vue,react}/dist`, so they are already covered by the existing `./components` entry — nothing needed there.

The line is for `packages/favicon/dist`, which sits outside `components/` alongside `packages/details-animation/dist` (listed here for the same reason). That one is not covered, and it matters: the job skips `yarn install` on a cache hit, so nothing rebuilds the workspace, and the package's `exports` point only at `dist/`. Without it the component spec cannot resolve `@cypress-design/favicon` on a cached run — which is exactly how it failed before this line was added.

.github/workflows/e2e-tests.yml
emilmilanov: Good catch, half of it. The component builds do land at `components/Favicon/{vue,react}/dist`, so they are already covered by the existing `./components` entry — nothing needed there.

The line is for `packages/favicon/dist`, which sits outside `components/` alongside `packages/details-animation/dist` (listed here for the same reason). That one is not covered, and it matters: the job skips `yarn install` on a cache hit, so nothing rebuilds the workspace, and the package's `exports` point only at `dist/`. Without it the component spec cannot resolve `@cypress-design/favicon` on a cached run — which is exactly how it failed before this line was added.

.github/workflows/template-test.yml
emilmilanov: Good catch, half of it. The component builds do land at `components/Favicon/{vue,react}/dist`, so they are already covered by the existing `./components` entry — nothing needed there.

The line is for `packages/favicon/dist`, which sits outside `components/` alongside `packages/details-animation/dist` (listed here for the same reason). That one is not covered, and it matters: the job skips `yarn install` on a cache hit, so nothing rebuilds the workspace, and the package's `exports` point only at `dist/`. Without it the component spec cannot resolve `@cypress-design/favicon` on a cached run — which is exactly how it failed before this line was added.

docs/public/apple-touch-icon.png
emilmilanov: Checked the pixels rather than the preview — the file has **zero** non-opaque pixels. Edges and corners are solid navy `#1B1E2E`, and it is encoded without an alpha channel at all (PNG colortype 2), which is deliberate: iOS composites this icon on white, so any transparency would show as a hard edge around the disc.

What you are seeing is the artwork's own geometry. The disc occupies columns 9–170 of 0–179, leaving a navy margin either side that reads as a gap against GitHub's dark preview background. Same margin top and bottom; it is only more noticeable left/right because the preview pane is wider than it is tall.

packages/favicon/assets/apple-touch-icon.png
emilmilanov: Checked the pixels rather than the preview — the file has **zero** non-opaque pixels. Edges and corners are solid navy `#1B1E2E`, and it is encoded without an alpha channel at all (PNG colortype 2), which is deliberate: iOS composites this icon on white, so any transparency would show as a hard edge around the disc.

What you are seeing is the artwork's own geometry. The disc occupies columns 9–170 of 0–179, leaving a navy margin either side that reads as a gap against GitHub's dark preview background. Same margin top and bottom; it is only more noticeable left/right because the preview pane is wider than it is tall.

emilmilanov: They are the same file, not two versions — `packages/favicon/assets/` is the source that ships in the npm package, and `docs/public/` is this site consuming it exactly the way any other site does, via the package's own `cypress-favicon <dir>` copy step. `build:docs` re-runs that copy, so the site cannot drift from what the package publishes, and it means the design system dogfoods its own package rather than special-casing itself.

They are byte-identical (verified by md5). The alternative is a build-time copy that is gitignored, which would leave the site unbuildable without the registry and hide the change from review.

docs/src/layouts/BaseLayout.astro
emilmilanov: Just this site. It is only there because we pass a `links` prop to drop the `manifest` tag — this site ships no `site.webmanifest`. A project using the defaults needs only:

```astro
import Favicon from '@cypress-design/vue-favicon'
<Favicon />
```

I have written that down in `components/Favicon/architecture.md` so it is not folklore.

docs/src/layouts/BaseLayout.astro
emilmilanov: Agreed — removed. That reasoning now lives in `components/Favicon/architecture.md`, next to the component, which is also where `.agents/pages/favicons.md` moved to.

AGENTS.md
emilmilanov: Moved it to `components/Favicon/architecture.md` instead — that is the existing convention for component implementation notes (per AGENTS.md: `components/<Name>/architecture.md`), and it puts the notes side by side with the component as you suggested on the other thread. The `pages/` entry in AGENTS.md is gone.

components/Favicon/architecture.md
emilmilanov: Done — it is now `components/Favicon/architecture.md`, next to the component rather than under `.agents/pages/`.

emilmilanov:
Thanks Ryan — all three addressed in 8d82d44, and I've replied inline on the rest.

**1. Which package to install.** There's now a table at the top of the page. Short version: everyone installs `@cypress-design/favicon` for the assets and the CLI; the two component packages are alternatives rather than a pair, each a build of the same component for one framework, and a site whose head isn't rendered by a component installs only the base package and calls a function. Installing one component does not give you both frameworks — that wasn't clear before and now it's stated explicitly.

**2. Product-neutral language.** Agreed, and done. Headings are now just `Astro`, `Docusaurus`, `EJS and other server-rendered HTML`, and the framework guidance is phrased as a property of your Astro config rather than of a particular Cypress site. No product or repo names remain in the page or the package ReadMe.

**3. Components section.** It's already there — adding `components/Favicon/` made `/components/Favicon` appear automatically with the framework tabs and `instructions.md`. So there are two pages now, and I've cross-linked them rather than merged: the component page covers props and per-framework install, and `/favicons` covers the asset set and the markup rules, which are broader than the component (Docusaurus and EJS consumers need that page but never touch a component).

Happy to fold `/favicons` into the component page entirely if you'd rather have one — it would mean the asset showcase becomes a demo under `docs/src/demos/`, and the rules-and-measurements prose moves into `instructions.md`. Your call.

One correction worth surfacing from the inline threads: the `apple-touch-icon.png` gaps aren't real. I decoded it — zero non-opaque pixels, no alpha channel at all. The disc occupies columns 9–170 of 180, so there's a navy margin that reads as a gap against GitHub's dark preview.

ryanjwilke:
@emilmilanov A couple things:

1. It does feel like the component documentation could have everything on it, even if you choose NOT to use the component. That's more of an escape hatch in my opinion.
2. Couldn't the `@cypress-design/vue-favicon` and `@cypress-design/react-favicon` packages include `@cypress-design/favicon` as a dependency so that way there's no need to have to figure out which one you use if you can cleanly use the icon? It's still a bit confusing that you need to install both packages to render a favicon. Or maybe you only install one package? That part just isn't quite as clear to me.

emilmilanov:
Both done in 500a7e2.

**1. Everything on the component page.** The standalone `/favicons` page is gone — `/components/Favicon` now carries it all. The asset showcase is the component demo, the framework tabs cover install and usage, and the rest lives in `instructions.md`: which package to install, picking a build, a *Without the component* section for Docusaurus and EJS as the escape hatch you described, the two omitted tags with their measurements, and how the assets are made.

**2. One install.** You were right, and it exposed a real gap. The component packages already depended on `@cypress-design/favicon`, so the assets and link data came along — but its `cypress-favicon` copy command didn't, because package managers only link commands from direct dependencies. So a site genuinely did need both. Each component package now ships its own `cypress-favicon` command and re-exports `FAVICON_LINKS`, so:

- Astro with Vue → install `@cypress-design/vue-favicon`, that's it
- Astro with React → install `@cypress-design/react-favicon`, that's it
- Docusaurus / EJS → install `@cypress-design/favicon`, that's it

Verified with packed tarballs rather than workspace links: installing only `vue-favicon` brings the base package, puts `cypress-favicon` on the path, and copies all six assets.

docs/src/layouts/BaseLayout.astro
```diff
@@ -30,14 +30,15 @@ const frontmatterTocHeadings = (Astro.props as { frontmatter?: { tocHeadings?: t
 const outlineHeadings = tocHeadings ?? frontmatterTocHeadings ?? headings
 const pageTitle = title ? `${title} — Cypress Design System` : 'Cypress Design System'
 const currentPath = Astro.url.pathname
+import Favicon, { FAVICON_LINKS } from '@cypress-design/react-favicon'
```
ryanjwilke: Just confirming on this one, but I believe most of the docs layout uses Vue components instead of React. We may want to be consistent on this component too. It's pretty minor though of course. I don't know the implications of using both frameworks in the same page, but I would imagine that we should stick with either Vue or React components so it doesn't require both JavaScript frameworks to render a page.

ryanjwilke (approved):
I approved, but I left you one minor comment to discuss with Claude Code to see what it thinks.

docs/src/layouts/BaseLayout.astro
emilmilanov: Good instinct, and I checked it rather than assumed — it turns out not to cost anything here.

**No React reaches the browser.** `react-favicon` has no `client:` directive, so Astro renders it to plain `<link>` tags at build time and ships no JavaScript for it. On `/components/Favicon` the only client runtime the page loads is `@astrojs/vue/dist/client.js`, for the sidebar `DocMenu` and the demo. So the page doesn't need both frameworks to render.

**Vue would be the consistent choice, but it breaks this site.** I did try `vue-favicon` in this layout first: every MDX page 500'd. The layout renders through MDX's JSX pipeline, which asks each registered renderer whether it owns the component, and `@astrojs/react` is registered first — its check reads `Component["$$typeof"].toString()` and throws on a Vue object instead of returning false. `DocMenu` avoids it only because it's `client:only="vue"`, which a favicon can't use since the tags must be in the initial HTML.

So React is the build that works in this site's head, at zero runtime cost. `vue-favicon` is the right one for cypress.io, which only registers Vue — verified there too.

If we ever drop React from this site's Astro config, switching to `vue-favicon` becomes a one-line change.

